Issue #10, opened 2023-06-16
Optimization Objective Skeleton Code


FireBoyAJ24:
### Purpose

The purpose of this tasklist is to organize the various optimization objectives that has been set in the [local-pathfinding design specification](https://ubcsailbot.atlassian.net/wiki/spaces/prjt22/pages/1763803137/Local+Pathfinding+Design+Specification).
[How Do We Calculate the Optimization Objectives?](https://ubcsailbot.atlassian.net/wiki/spaces/prjt22/pages/1860403201/How+Do+We+Calculate+the+Optimization+Objectives)

### Tasks

- [x]  Set up a basic framework which outputs a test optimization value
- [x] Distance and upwind/downwind objectives
- [x] #25
- [x] #26

### Resources
- Initialize simple setup object: https://github.com/UBCSailbot/raye-local-pathfinding/blob/master/python/updated_geometric_planner.py#L164
- Objectives: https://github.com/UBCSailbot/raye-local-pathfinding/blob/master/python/planner_helpers.py
- Boat state example: https://github.com/UBCSailbot/raye-local-pathfinding/blob/master/python/Sailbot.py#L19

Issue #3, opened 2023-06-08
Write tests


patrick-5546:
### Description

Setup test infrastructure and write initial tests while the codebase is still small.

### Resources

- Test framework: https://docs.pytest.org/en/7.1.x/getting-started.html
- Old codebase: https://github.com/UBCSailbot/raye-local-pathfinding/tree/master/test
    - Doesn't use pytest but still good to reference

PR #13, opened 2023-06-20
Add path states

patrick-5546:
### Description

- Needed for #3 
- Helps with #10 

Abstract the state that `LocalPath` and `OMPLPath` need into their own classes. This has a number of benefits:

- Their functions' arguments don't have to change as much
- So that we can easily create states in our tests (TBD)
- Reduce the dependencies between `Sailbot`, `LocalPath`, and `OMPLPath`

### Verification

- [x] Runs on my computer

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,8 +12,8 @@ Using main launch file: `ros2 launch local_pathfinding main_launch.py`
 
 Launch arguments are added to the run command in the format `<name>:=<value>`.
 
-| name | description | value |
-| - | - | - |
-| `log_level` | Logging level | A [severity level][severity level] |
+| name        | description   | value                                                 |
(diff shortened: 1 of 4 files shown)


patrick-5546:
This is needed to type hint without cyclic imports (files import each other):
https://github.com/UBCSailbot/local_pathfinding/blob/5f07059f68c2571a1c0ef161bfb82981d23338a6/local_pathfinding/ompl_path.py#L17-L18
For more details, you can see this stack overflow post and the most upvoted answer: https://stackoverflow.com/questions/39740632/python-type-hinting-without-cyclic-imports

FireBoyAJ24 (approved):
If the primary purpose of this pull request is to just setup the preliminary connection between the messages from the ROS to the local pathfinding logic, then I approve.

PR #14, opened 2023-06-23
Write tests

### LINKED ISSUE #3: Write tests

patrick-5546:
### Description

- Resolves #3 

Setup test infrastructure:
- Tests standalone functionality: not all functions are tested (e.g., interfaces, dataflows may not be tested)
- Each file has a corresponding test file
    - No tests for ROS node files because you can't run them standalone
        - ROS node files should be kept small

### Verification

- [x] Tests pass

### DIFF
--- a/local_pathfinding/local_path.py
+++ b/local_pathfinding/local_path.py
@@ -14,12 +14,18 @@ class LocalPathState:
         global_path: GlobalPath,
         filtered_wind_sensor: WindSensor,
     ):
-        if gps:  # TODO: remove when mocks can be run
+        if gps:  # TODO: remove when mock can be run
             self.position = gps.lat_lon
             self.speed = gps.speed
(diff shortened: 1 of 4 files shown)


patrick-5546:
This is an example of parameterizing a test in pytest:
https://github.com/UBCSailbot/local_pathfinding/blob/b1829fe340add9254f2a902325b4fcb752ce6bc4/test/test_ompl_path.py#L46-L53
For more details, see https://docs.pytest.org/en/7.3.x/how-to/parametrize.html

PR #30, opened 2023-10-03
Add coordinate system functions and classes

patrick-5546:
### Description

- Relevant exploration notebook: https://github.com/UBCSailbot/notebooks/blob/main/local_pathfinding/coord_systems.ipynb


patrick-5546:
@SPDonaghy FYI once this is merged in you should update your main branch and merge it with your development branch so that you can use these functions and classes. You will also need to update sailbot workspace to get the dependencies

Issue #25, opened 2023-07-31
Create a Heading Objective


FireBoyAJ24:
1. What is defined as "better"? A better objective method is one where it improves
one or all of the following criteria:
    a. It can produce smaller tacks
    b. There are no issues that can lead to the Sailbot to get damaged when there is an obvious safer path.
2. We have two methods to test:
    a. Is it better to measure the angle from the previous position (Raye's method)
    b. Is it better to measure the angle from the final position (A heuristic, my method)
 
All the test process, data, results and evaluation will be rendered in confluence and will be separated in files as reference.

Further Information is in ["How Do We Calculate the Optimization Objectives?"](https://ubcsailbot.atlassian.net/wiki/spaces/prjt22/pages/1860403201/How+Do+We+Calculate+the+Optimization+Objectives)

PR #32, opened 2023-10-07
25 create a heading objective

### LINKED ISSUE #24: Create multiple tests for the different Distance optimizers
1. How do we test the distance optimizers?
    a. See how each of the methods work
    b. We are looking for the differences of the distance between two points for all
the selected methods. The points distance will range between 1m, 10m, 100m,
1km, 5km, 10km, 50km and 100km.
2. Methods that we are going to test
    a. Test the OMPL's distance objective (Raye's method)
    b. Test the Euclidean Distance method
    c. Test the Thunderbird's distance method
    d. Check and test whether Manhattan or Minkowski Distance
    
All the test process, data, results and evaluation will be rendered in confluence and will be separated in files as reference.
    
Further Information is in ["How Do We Calculate the Optimization Objectives?"](https://ubcsailbot.atlassian.net/wiki/spaces/prjt22/pages/1860403201/How+Do+We+Calculate+the+Optimization+Objectives)
https://ompl.kavrakilab.org/optimizationObjectivesTutorial.html

FireBoyAJ24:
### Description

- #25 
- #24 

### Resolves #24 and #25

I have added the distance, minimum turning and upwind/downwind optimization objectives. We can use this as a starting point to build upon, test and verify our optimization objectives.

Note: Test cases have not been made as requested in #24 since that is a separate major issue to tackle.

### Verification

- [x] Check whether all the distance objective functions are not producing an error.
- [x] Check whether all the minimum turning objective functions are not producing an error.
- [x] Check whether the wind objective function is not producing an error

### Resources

- https://ubcsailbot.atlassian.net/wiki/spaces/prjt22/pages/1860403201/How+Do+We+Calculate+the+Optimization+Objectives


FireBoyAJ24:
I got two errors I need to resolve.

1. When I run the ompl path optimization objective, why do I get the following error.
image
image

2. When I run goalPathTurnCost I get the following error
image
image

PR #34, opened 2023-10-09
Create boat class

### LINKED ISSUE #29: Create boat obstacle
### Purpose

Avoid boats

### Description

Create a `Boat` class that stores information that we get from the ROS network ([HelperAISSHIP.msg](https://github.com/UBCSailbot/custom_interfaces/blob/main/msg/HelperAISShip.msg)) and has an `isValid` function that returns whether a given xy coordinate is where we think the boat will be.

### Resources

- Files from Raye's codebase
    - state validity checker: https://github.com/UBCSailbot/raye-local-pathfinding/blob/master/python/planner_helpers.py#L35
    - obstacles: https://github.com/UBCSailbot/raye-local-pathfinding/blob/master/python/obstacles.py
- @patrick-5546 will add latlon <-> xy and other prerequisite functions/classes in UBCSailbot/local_pathfinding#30 by the October 5 meeting

SPDonaghy:
### Description
- Resolves #29 

1. A module titled obstacles.py was created
2. An obstacle Interface was created for later functionality
3. A Boat class was implemented, as well as an is_valid function
4. A test file titled test_obstacles.py was created to test the Boat init function and visualize a boat's collision cone.

### Verification

1. Tested the Boat init function by creating a boat, and verifying its dimensions, position, and orientation with a plotly scatter plot
2. Tested the is_valid function by passing points that are inside and outside of the boat's collision cone.


local_pathfinding/obstacles.py
```diff
@@ -0,0 +1,152 @@
...
+        # This factor can be adjusted to change the scope/width of the collision cone
+        COLLISION_CONE_STRETCH_FACTOR = 1.3
+
+        speed_mps = knots_to_meters_per_second(speed)
+
+        # coordinates of the center of the boat
+        x, y = position[0], position[1]
+
+        # Points of the boat collision box polygon before rotation and centred at the origin
+        points = np.array(
```
SPDonaghy: I would like to receive feedback on how to shape the boat's collision box.  I am defining the collision box as the region where the boat may possibly be, over a certain amount of time: delta_time. 

This array contains the set of points which describe the boat's collision box.

Right now, it is shaped like a trapezoid. with the narrower end being where the boat currently is, and the wider end extends a certain distance ahead depending on the boat's length, speed, and delta-time (which are all parameters passed to the Boat init function.

I chose this shape so that, if the boat is not moving, its hull is still completely contained in the collision box.

But I am also considering defining the collision box as follows: 

The ship is inscribed in a circle, and a cone shape extends from it, in the direction of the course over ground (track made good). This shape will ensure that even if the ship's heading does not match its COG then no points on the boat will stick out of the collision box. This can happen with the simpler trapezoid shape

local_pathfinding/obstacles.py
```diff
@@ -0,0 +1,152 @@
+import math
+
+import numpy as np
+from shapely.geometry import Polygon
+
+
+# TODO Look into adding more data available from AIS
```
SPDonaghy: Rate of turn might be a useful data value to better depict the expected path of boats

local_pathfinding/obstacles.py
```diff
@@ -0,0 +1,152 @@
...
+            -course_over_ground (float, degrees): COG of the boat, clockwise from true north
+        """
+        self.id = id
+        self.speed = speed
+
+        self.collision_cone = self.create_collision_cone(
+            width, length, position, speed, delta_time, course_over_ground
+        )
+
+    def create_collision_cone(
```
SPDonaghy: Later on, I would like to see how I can reuse this function (or the entire class) to create land objects from geo data.

test/test_obstacles.py
```diff
@@ -0,0 +1,29 @@
+import numpy as np
+import plotly.graph_objects as go
```
SPDonaghy: I think Plotly will be essential in visualizing what I am working on with obstacles, moving forward. However, it runs very slowly sometimes, I tried the commands for setting the hostname and that does improve it, but I am finding I need to do that every time I start the codespace.

local_pathfinding/obstacles.py
```diff
@@ -0,0 +1,152 @@
...
+        To avoid confusion, the physical dimensions, position, and COG of the boat are
+        stored in the aggregated Polygon object, not the Boat object itself.
+        So that all we need to do to update the boat's position or COG is to update the
+        polygon's position, size, and rotation.
+
+        points (x,y) are measured in meters, with the origin set at the current waypoint.
+
+    """
+
+    def __init__(self, id, width, length, position, speed, delta_time, course_over_ground):
```
patrick-5546: Arguments should include https://github.com/UBCSailbot/custom_interfaces/blob/main/msg/HelperAISShip.msg and any other information needed. Thus latlon to xy conversion should be done inside this, and you will need to pass in the reference point (next global waypoint)

Example of importing msgs in codebase:
https://github.com/UBCSailbot/local_pathfinding/blob/a3dc649ab4cc077ed3a4a0e7ef7b23b93d8b8a92/local_pathfinding/node_navigate.py#L4

local_pathfinding/obstacles.py
```diff
@@ -0,0 +1,152 @@
...
+        """
+        Args:
+
+            -id (int): MMSI number of the boat
+            -width (float): width of the boat in meters
+            -length (float): length of the boat in meters
+            -position (XY): x,y coordinates of the boat converted from lat/lon
+            -speed (float): speed of the boat in knots
+            -delta_time (float): time in seconds, to calculate the boat's expected position
```
patrick-5546: Rather than time it should use the current position and maximum speed of sailbot

local_pathfinding/obstacles.py
```diff
@@ -0,0 +1,152 @@
...
+        """
+        Checks if a point is within the boat's collision cone
+
+        Args:
+            point (Point): a shapely Point object representing the point to be checked
+
+        Returns:
+            bool: True if the point is not within the boat's collision cone
+        """
+        return not self.collision_cone.contains(point)
```
patrick-5546: This should be able to be defined inside `ObstacleInterface`

local_pathfinding/obstacles.py
```diff
@@ -0,0 +1,152 @@
+import math
+
+import numpy as np
+from shapely.geometry import Polygon
+
+
+# TODO Look into adding more data available from AIS
+# TODO Decide between using COG and SOG or Heading and Speed
+class ObstacleInterface:
```
patrick-5546: Rename to `Obstacle`

test/test_obstacles.py
```diff
@@ -0,0 +1,29 @@
...
+y = np.array(boat1.collision_cone.exterior.coords.xy[1])
+
+# Create a Plotly figure for boat1's collision cone with a hole representing the hull
+fig1 = go.Figure(go.Scatter(x=x, y=y, fill="toself"))
+
+fig1.update_layout(yaxis_range=[-200, 200], xaxis_range=[-200, 750])
+
+# check the is_valid function
+assert not boat1.is_valid(Point(100, 50))
+assert boat1.is_valid(Point(100, 100))
```
patrick-5546: Wrap in test method and parameterize using pytest:
- Example in codebase
https://github.com/UBCSailbot/local_pathfinding/blob/a3dc649ab4cc077ed3a4a0e7ef7b23b93d8b8a92/test/test_ompl_path.py#L46-L60
- Pytest documentation: https://docs.pytest.org/en/7.4.x/how-to/parametrize.html

test/test_obstacles.py
```diff
@@ -0,0 +1,29 @@
...
+boat1 = Boat(999, 26, 52, position, 30, 10, 70)
+
+# Extract exterior coordinates for boat1's collision cone
+x = np.array(boat1.collision_cone.exterior.coords.xy[0])
+y = np.array(boat1.collision_cone.exterior.coords.xy[1])
+
+# Create a Plotly figure for boat1's collision cone with a hole representing the hull
+fig1 = go.Figure(go.Scatter(x=x, y=y, fill="toself"))
+
+fig1.update_layout(yaxis_range=[-200, 200], xaxis_range=[-200, 750])
```
patrick-5546: Put the plotly stuff (manual inspection) at the end and only run when running the file
```py
if __name__ == "__main__":
    import plotly.graph_objects as go
    ...
```

local_pathfinding/obstacles.py
```diff
@@ -0,0 +1,152 @@
...
+        """
+        self.id = id
+        self.speed = speed
+
+        self.collision_cone = self.create_collision_cone(
+            width, length, position, speed, delta_time, course_over_ground
+        )
+
+    def create_collision_cone(
+        self, width, length, position, speed, delta_time, course_over_ground
```
patrick-5546: Add type hints to all functions

patrick-5546 (commented):
Great start! Here are a couple things I noticed:

SPDonaghy:
I believe it is just failing colcon because my changes to custom_interfaces are not merged to main. I have submitted a PR for those changes.

local_pathfinding/obstacles.py
```diff
@@ -0,0 +1,340 @@
...
+
+class Obstacle:
+
+    """
+    This Interface describes general obstacle objects which are
+    anything which the sailbot must avoid.
+
+    The choice was made to create an interface rather than a class as Boat and LandMass objects
+    may be very different objects which share an "avoidable" property and a position, but not much
+    else.
```
patrick-5546: I don't think this is true anymore

local_pathfinding/obstacles.py
```diff
@@ -0,0 +1,340 @@
...
+        self,
+        reference: LatLon,
+        sailbot_position: LatLon,
+        sailbot_speed: float,
+    ):
+        """
+        Args:
+            -reference (LatLon): latitude and longitude of the next global waypoint
+            -sailbot_position (LatLon): lat and lon position of SailBot
+            -sailbot_speed (float): speed of the SailBot in km/h
```
patrick-5546: I don't think we need a docstring for the init function in this case. However, let's move this section to the class attributes.

For example
https://github.com/UBCSailbot/local_pathfinding/blob/bd5a710ba1af486d615790a4b2f9c7f94dc87372/local_pathfinding/local_path.py#L53-L61

patrick-5546: The only difference is that `sailbot_position` would be type `XY` instead of `LatLon`

local_pathfinding/obstacles.py
```diff
@@ -0,0 +1,340 @@
...
+            - Maritime Mobile Service Identity (9 digit int)
+            - latitude (float) latitude in degrees
+            - longitude (float) longitude in degrees
+            - speed over ground (float) speed in knots over ground
+            - course over ground (float) degrees measured clockwise from true north
+            - boat dimensions (float) width and length of the boat in meters
+            - rate of turn (float) ROT in AISROT scale -126 to +126 corresponding to
+                -708 to +708 deg/min
+
+            - other information is also available but not used here
```
patrick-5546: Scope: shouldn't be documented here. Docstrings should only include what is necessary to understand the implementation. Can you move this to `HelperAISShip.msg`

local_pathfinding/obstacles.py
```diff
@@ -0,0 +1,340 @@
...
+            -sailbot_speed (float): speed of the SailBot in km/h
+            -ais_ship (HelperAISShip): an AISShip object, containing the following information:
+                -id (int): MMSI number of the boat
+                -width (float): width of the boat in meters
+                -length (float): length of the boat in meters
+                -lat_lon (LatLon): latitude and longitude of the boat
+                -speed_over_ground (float): speed of the boat in knots, over ground
+                -course_over_ground (float): COG of the boat, in degrees, clockwise from true north
+                -rate_of_turn (float): ROT of the boat in AISROT scale -126 to +126 corresponding
+                    to -708 to +708 degrees per minute
```
patrick-5546: Again no need to document here, people can refer to the definition of `HelperAISShip.msg`

local_pathfinding/obstacles.py
```diff
@@ -0,0 +1,340 @@
...
+    """
+    This class describes boat objects which Sailbot must avoid.
+    Also referred to target ships or boat obstacles.
+
+
+    Attributes
+    ----------
+        -id (int): MMSI number of the boat
+        -speed_over_ground (float): speed of the boat, over ground, in knots
+        -collision_cone (Polygon): a shapely Polygon object representing the boat's collision box
```
patrick-5546: This should contain only attributes that are unique to `Boat`. Should move `collision_cone` to `Obstacle` attributes

local_pathfinding/obstacles.py
```diff
@@ -0,0 +1,340 @@
...
+    """
+
+    def __init__(
+        self,
+        reference: LatLon,
+        sailbot_position: LatLon,
+        sailbot_speed: float,
+        ais_ship: HelperAISShip,
+    ):
+        """
```
patrick-5546: imo no need to add docstring to init function unless you're doing something crazy

local_pathfinding/obstacles.py
```diff
@@ -0,0 +1,340 @@
...
+        sailbot_speed: float,
+    ) -> Polygon:
+        """
+        Creates a Shapely Polygon to represent the boat's collision_cone, which is sized,
+        orientated and positioned according to the boat's COG, SOG, and position.
+
+        The polygon is oversized according to the collision zone safety buffer, for
+        added assurrance that the boat will be entirely contained by the polygon.
+
+        Args:
```
patrick-5546: Missing `sailbot_speed`

local_pathfinding/obstacles.py
```diff
@@ -0,0 +1,340 @@
...
+            position,
+            course_over_ground,
+            speed_over_ground_kmph,
+            sailbot_position,
+            sailbot_speed,
+        )
+
+        # This factor can be adjusted to change the scope/width of the collision cone
+        # TODO This feels too arbitrary, maybe will incorporate ROT at a later time
+        COLLISION_CONE_STRETCH_FACTOR = projected_distance * 1.5
```
patrick-5546: All caps reserved for constants defined at the top of the file. Can make `1.5` a constant

local_pathfinding/obstacles.py
```diff
@@ -0,0 +1,340 @@
...
+        The polygon is oversized according to the collision zone safety buffer, for
+        added assurrance that the boat will be entirely contained by the polygon.
+
+        Args:
+            width (float): width of the boat in meters
+            length (float): length of the boat in meters
+            position (XY): x,y coordinates of the boat in km
+            speed_over_ground (float): speed of the boat in knots, over ground
+            course_over_ground (float): COG of the boat in degrees clockwise from true north
+            rate_of_turn (float): ROT of the boat in AISROT scale -126 to +126 corresponding to
```
patrick-5546: rate of turn unused

local_pathfinding/obstacles.py
```diff
@@ -0,0 +1,340 @@
...
+        if (2 * (v1 * (a - c) - v2 * (b - d))) ** 2 - 4 * (
+            v1**2 + v2**2 - (sailbot_speed**2)
+        ) * (a - c) ** 2 + (b - d) ** 2 <= 0:
+            return -1
+
+        # The solution to the quadratic formula is the time until the boats collide
+        t = np.roots(quadratic_coefficients)
+
+        # Return the smaller positive time
+        return min([i for i in t if i > 0])
```
patrick-5546: Is it possible that there could be no element > 0? If so, `min` would raise a `ValueError`

local_pathfinding/obstacles.py
```diff
@@ -0,0 +1,340 @@
...
+        """
+        self.reference = reference
+
+        # Position of the Sailbot in XY
+        self.sailbot_position = latlon_to_xy(reference, sailbot_position)
+
+        self.sailbot_speed = sailbot_speed
+
+        # This should be defined in child classes
+        self.collision_zone = Polygon(None, None)
```
patrick-5546: What do you think about this for debugging purposes
- Set `self.collision_zone = None` here
- In `is_valid()`
  ```py
  if self.collision_zone is None:
    raise ValueError("Collision zone has not been initialized")
  ```

patrick-5546 (commented):
Great work on this PR! I am really impressed by your code quality and readability. I aligned the docstring formatting with the rest of the codebase and noted a couple small things.

local_pathfinding/obstacles.py
patrick-5546: Can remove it until/if it is used?

local_pathfinding/obstacles.py
patrick-5546: If you implement this, you will probably want to add the corresponding test

local_pathfinding/obstacles.py
SPDonaghy: Yes, that is possible. I will add a try except block to check for that and return -1 if there are no positive elements. 

Generally, its ok if there doesn't end up being any positive times, just means they cant collide with the current parameters. So projecting the Boat's position up to the maximum distance would be the appropriate action in that case.

local_pathfinding/obstacles.py
SPDonaghy: So sailbot_position will be published as an XY correct?
Will reference also be published as XY? 

I'm a little confused on what will be XY and what will be LatLon, I assumed I needed to use latlon_to_xy in my functions, because locations will be published in LatLon, passed in to functions as LatLon, and then converted to XY inside my functions.

patrick-5546: Hopefully we can keep the XY coordinate system internal to this file (not thought about anywhere else). The `self.sailbot_position` class attribute would be type XY. The reference XY coordinate would be (0, 0)

local_pathfinding/obstacles.py
SPDonaghy: I like this idea, I will implement it.

patrick-5546: You will also want to update the docstring

```py
"""

Raises:
    ValueError: When collision zone has not been initialized.
"""
```

patrick-5546 (approved):
I think this PR is ready to merge, great job! You'll have to merge in your custom_interfaces PR first and rerun failed tests.

Notes about my nitpicks:
- Add docstring for any exceptions raised
- `collision_zone` is defined in `Obstacle`, so to respect that in `Boat` I call an Obstacle function that updates the collision zone
- Class attributes are any instance variables you define

What would be good to work on next is using this in the other files to add obstacle avoidance.

patrick-5546:
~~Also it would be great if you presented your work at the meeting on Saturday to show everyone what you've been up to!~~ actually not done yet lol

local_pathfinding/obstacles.py
```diff
@@ -0,0 +1,267 @@
...
+        self.id = ais_ship.id
+
+        position = latlon_to_xy(
+            self.reference, LatLon(ais_ship.lat_lon.latitude, ais_ship.lat_lon.longitude)
+        )
+        self.position = position
+
+        collision_zone = self.create_collision_cone(
+            ais_ship.dimensions.width,
+            ais_ship.dimensions.length,
```
patrick-5546: You can store all these arguments as instance variables so that you don't have to pass them to each function

local_pathfinding/obstacles.py
```diff
@@ -0,0 +1,267 @@
...
+            position,
+            ais_ship.sog.sog,
+            ais_ship.cog.cog,
+            self.sailbot_position,
+            self.sailbot_speed,
```
patrick-5546: Use the instance variables inside the function definition rather than passing it in as an argument

local_pathfinding/obstacles.py
```diff
@@ -0,0 +1,267 @@
...
+        width = meters_to_km(width)
+        length = meters_to_km(length)
+
+        # Calculate distance the boat will travel before soonest possible collision with Sailbot
+        projected_distance = self.calculate_projected_distance(
+            position,
+            course_over_ground,
+            speed_over_ground,
+            sailbot_position,
+            sailbot_speed,
```
patrick-5546: Same for these

local_pathfinding/obstacles.py
```diff
@@ -0,0 +1,267 @@
...
+        # Speed over ground vector of the boat obstacle
+        boat_sog_vector = np.array(
+            [
+                speed_over_ground_kmph * np.sin(math.radians(course_over_ground)),
+                speed_over_ground_kmph * np.cos(math.radians(course_over_ground)),
+            ]
+        )
+
+        time_to_intersection = self.calculate_time_to_intersection(
+            position, boat_sog_vector, sailbot_position, sailbot_speed
```
patrick-5546: Same for these

test/test_obstacles.py
patrick-5546, line 1: Do you think it would be possible to add more unit tests for the boat obstacle? Like tests for create_collision_cone, calculate_projected_distance, and calculate_time_to_intersection

local_pathfinding/coord_systems.py
```diff
@@ -45,6 +45,22 @@ def cartesian_to_true_bearing(cartesian: float) -> float:
     return (90 - cartesian + 360) % 360
 
 
+def knots_to_kilometers_per_hour(knots: float) -> float:
```
patrick-5546: Although trivial, we should add tests for each function

patrick-5546 (changes requested):
Sorry thought of some more things to do in this PR

local_pathfinding/coord_systems.py
patrick-5546: Also is this function even used anywhere?

test/test_obstacles.py
patrick-5546: If possible the test arguments should make sense intuitively (not always require calculation). If it does require calculation it would be helpful to add the function that calculates it.

local_pathfinding/coord_systems.py
SPDonaghy: No, it used to be. But if speeds can be published in kmph its not needed

SPDonaghy:
**dynamic variables:** 
- sog
- cog
- position
- sailbot_position
- sailbot_speed

**non-dynamic variables:**
- id
- width
- length

Hi Patrick, I am just resolving some conflicts, with my most recent changes and your commit. (Lesson learned I should have pulled before making more changes)

Having update_collision_zone(Polygon) as a non static method makes sense to me. It seems like the intent is that we can just call this method to update the dynamic variables instead of making a new boat object every time something changes.

I am just wondering for create_collision_cone() and its helper functions (proj distance and proj time) do we want these to be non-static as well? Is it okay for them to still have parameters for the dynamic variables listed above? And then just use the non dynamic instance variables (self.width and self.length) inside these functions?

I interpreted your previous comments about using instance variables to mean we shouldn't pass anything to these functions as parameters. But that would require storing the dynamic variables as instance variables. Which is something I want to avoid, so that we don't have that data stored in the instance and in the instance's collision_zone.

patrick-5546:
> Lesson learned I should have pulled before making more changes

Yes this is something we all needed to go through, but glad you did sooner rather than later lol

> I am just wondering for create_collision_cone() and its helper functions (proj distance and proj time) do we want these to be non-static as well? Is it okay for them to still have parameters for the dynamic variables listed above? And then just use the non dynamic instance variables (self.width and self.length) inside these functions?

You make a good point. Rather than a direct answer, I will give some more background about the usecases of the Boat class. Primarily it is to feed into our path planner for collision avoidance. But our second usecase is to visualize boat's and their predicted positions on the website. Thus we will want to ensure that the collision zone that we are publishing was computed using the dynamic information that we are publishing. One way to do this is to store the dynamic variables as instance variables. along with the collision zone. Side note: it might be helpful to have the time of the information as well.

For `create_collision_cone()`, we can pass in an argument of type `HelperAISShip`. Then it would update the dynamic instance variables and create the polygon. May be good to raise an exception if the id of the argument doesn't match `self.id`. Also infrastructure wise, one idea is to overload `create_collision_zone()`: rename `create_collision_cone()` to `create_collision_zone(HelperAISShip)` and make its last line `self.create_collision_zone(Polygon)`.

For proj distance and proj time, we could make these class variables (does not use instance variables, `@staticmethod`). This would make them easier to test separately.

> But that would require storing the dynamic variables as instance variables. Which is something I want to avoid, so that we don't have that data stored in the instance and in the instance's collision_zone.

Not sure if I understand this, how are the dynamic variables stored in the collision zone?

SPDonaghy:
Thanks for the comments! I'll get working on this again tonight or tomorrow. 

> Not sure if I understand this, how are the dynamic variables stored in the collision zone?

No you're correct, they are used to make the collision zone, not really stored in it.

SPDonaghy:
I realized, along with functions for updating a boat's collision zone periodically, The Obstacle class needs a function to update dynamic information about Sailbot (position and speed) periodically as well. I'll write something up later today.

Also I saw your comment about including timestamps in AISShip messages as well. As I understand, it won't really be needed in this module for anything. And maybe I can just make a new pr in custom_interfaces at some point, to add that?

Also, I still want to explore how missing/unavailable data should be handled. (Ex. I've seen A COG of 501 means unavailable)

patrick-5546:
> Also, I still want to explore how missing/unavailable data should be handled. (Ex. I've seen A COG of 501 means unavailable)

This should be handled by NET, @hhenry01 FYI

local_pathfinding/obstacles.py
```diff
@@ -0,0 +1,254 @@
...
+        """
+        if self.collision_zone is None:
+            raise ValueError("Collision zone has not been initialized")
+
+        # contains() requires a shapely Point object as an argument
+        point = Point(*point)
+
+        return not self.collision_zone.contains(point)
+
+    def update_collision_zone(self, collision_zone: Polygon):
```
patrick-5546: What do you think about adding what is in `create_collision_zone` here and removing it? This would reduce the number of time we go back and forth between `Boat` and `Obstacle`

local_pathfinding/obstacles.py
```diff
@@ -0,0 +1,254 @@
...
+
+
+class Boat(Obstacle):
+    """Describes boat objects which Sailbot must avoid.
+    Also referred to target ships or boat obstacles.
+
+    Attributes:
+        id (int): MMSI ID of the boat.
+        width (float): Width of the boat in meters.
+        length (float): Length of the boat in meters.
```
patrick-5546: Only attribute now is `ais_ship`

local_pathfinding/obstacles.py
```diff
@@ -0,0 +1,254 @@
...
+        sailbot_position: LatLon,
+        sailbot_speed: float,
+        ais_ship: HelperAISShip,
+    ):
+        super().__init__(reference, sailbot_position, sailbot_speed)
+
+        self.ais_ship = ais_ship
+        self.update_collision_zone(self.create_boat_collision_zone(self.ais_ship))
+
+    def create_boat_collision_zone(self, ais_ship: HelperAISShip) -> Polygon:
```
patrick-5546: How about we make this `update_boat_collision_zone`? Then we can call it directly in init. The last line of this function would call `self.update_collision_zone`

local_pathfinding/obstacles.py
```diff
@@ -0,0 +1,254 @@
...
+
+        Returns:
+            Polygon: Shapely Polygon representing the boat's collision zone.
+        """
+        if self.ais_ship.id != ais_ship.id:
+            raise ValueError("Argument AIS Ship ID does not match this Boat instance's ID")
+
+        # Ensure the instance variable ais_ship is the most up-to-date version
+        # Because this function may be called outside of the constructor
+        self.ais_ship = ais_ship
```
patrick-5546: One way to resolve this redundancy is to make the `ais_ship` argument optional

local_pathfinding/obstacles.py
```diff
@@ -0,0 +1,254 @@
...
+                [collision_zone_stretch * width, length / 2 + projected_distance],
+                [width / 2, -length / 2],
+            ]
+        )
+
+        collision_zone_poly = af.rotate(collision_zone_poly, -cog, origin=(0, 0))
+
+        return Obstacle.create_collision_zone(collision_zone_poly, position)
+
+    def calculate_projected_distance(self, ais_ship: HelperAISShip) -> float:
```
patrick-5546: If you pass in the boat attributes needed directly you can make this static

local_pathfinding/obstacles.py
```diff
@@ -0,0 +1,254 @@
...
+
+        t = self.calculate_time_to_intersection(ais_ship)
+
+        if t < 0:
+            # Sailbot and this Boat will never collide
+            return PROJ_TIME_NO_COLLISION * ais_ship.sog.speed
+
+        return t * ais_ship.sog.speed
+
+    def calculate_time_to_intersection(self, ais_ship: HelperAISShip) -> float:
```
patrick-5546: If you pass in the boat attributes needed directly you can make this static

local_pathfinding/obstacles.py
patrick-5546: Why not add implementation in `calculate_time_to_intersection` here? This function is pretty small, but ultimately utu just a suggestion

local_pathfinding/obstacles.py
```diff
@@ -0,0 +1,254 @@
...
+
+        Args:
+            reference (LatLon): LatLon position of the updated global waypoint.
+        """
+        self.reference = reference
+        self.sailbot_position = latlon_to_xy(self.reference, self.sailbot_position_latlon)
+
+        if isinstance(self, Boat):
+            # regenerate collision zone with updated reference point
+            self.update_collision_zone(self.create_boat_collision_zone(self.ais_ship))
```
patrick-5546: Can replace with just `self.update_boat_collision_zone` if following suggestions above

patrick-5546 (commented):
I really like the changes you made! I think these suggestions should be the last things from me.

local_pathfinding/test.txt
patrick-5546, line 1: Was this committed by mistake?

patrick-5546 (approved):
Looks good to me. Would be interested in diving deeper into the math at some point, perhaps in a SW design review.

I changed `calculate_projected_distance()` back to an instance function because I saw that all the arguments that it was being passed were instance variables.

Also for performance reasons we would like to do as little math in raw Python as possible (within reason), instead using `math` for single numbers and `numpy` for arrays. Can any of the raw math that you did in `update_collision_zone()` or `calculate_projected_distance()` be converted to use these libraries? And for calculations that are performed multiple times, consider using a variable

test/test_obstacles.py
```diff
@@ -0,0 +1,463 @@
...
+
+    # There is some error in the latlon_to_xy conversion but the results are close
+    assert translation == pytest.approx(displacement, rel=0.1), "incorrect translation"
+
+
+if __name__ == "__main__":
+    """VISUAL TESTS
+
+    The collision zone length can be verified visually, using the plotly chart below.
+    """
```
patrick-5546: Would be helpful to add what to check to verify that everything is working

SPDonaghy:
I made a few optimization improvements:

-  storing repeatedly calculated/retrieved values in variables
- importing specific functions from libraries to avoid using dot operator or importing entire libraries
- reduced the number of polygon objects created during updates by combining rotation and translation into a single transform.

I wasn't sure though, how to incorporate math to make any of my calculations not run in python. Would you be able to explain more what you mean by using math for single numbers?

patrick-5546:
I think what you have is good. I reverted some changes because although they're technically faster, they aren't a bottleneck and make the code less readable and concise

PR #37, opened 2023-10-14
Akshanjay/optimization objective skeleton code

FireBoyAJ24:
### Description
Made the optimization objective skeleton code. I have added three optimization objective skeleton codes which just raise not implemented errors. I have added an RRT planner just for testing the optimization objectives. 

### Verification

- [x] Check whether there are no output errors from ROS
- [x] Check whether the code passes all the test cases

### Resources

- #10


local_pathfinding/objectives.py
```diff
@@ -0,0 +1,101 @@
...
+
+from ompl import base as ob
+
+# Upwind downwind cost multipliers
+UPWIND_MULTIPLIER = 3000.0
+DOWNWIND_MULTIPLIER = 3000.0
+
+# Upwind downwind constants
+UPWIND_MAX_ANGLE_DEGREES = 40.0
+DOWNWIND_MAX_ANGLE_DEGREES = 20.0
```
patrick-5546: These are unused right now, should we remove them?

local_pathfinding/objectives.py
```diff
@@ -0,0 +1,101 @@
...
+
+
+class Objective(ob.StateCostIntegralObjective):
+    """All of our optimization objectives inherit from this class.
```
patrick-5546, line 7: This is to enforce common behavior across our objectives. Currently it enforces that:
- All our objectives implement motion cost
- It says nothing about whether it implements a state cost
- All our objectives enable motion cost interpolation
Does this align with what you were going for?

local_pathfinding/objectives.py
```diff
@@ -0,0 +1,101 @@
...
+    Notes:
+    - This class inherits from the OMPL class StateCostIntegralObjective:
+        https://ompl.kavrakilab.org/classompl_1_1base_1_1StateCostIntegralObjective.html
+    - Camelcase is used for functions that override OMPL functions, as that is their convention.
+
+    Attributes:
+        space_information: https://ompl.kavrakilab.org/classompl_1_1base_1_1SpaceInformation.html
+    """
+
+    def __init__(self, space_information):
```
patrick-5546, line 18: I'm unsure of the type of space_information, if you find out you can add it (optional)

local_pathfinding/objectives.py
```diff
@@ -0,0 +1,101 @@
...
+
+
+class DistanceObjective(Objective):
+    def __init__(self, space_information):
+        super().__init__(space_information)
+
+    def motionCost(self, s1: ob.SE2StateSpace, s2: ob.SE2StateSpace) -> ob.Cost:
+        return ob.Cost()
+
+    def _get_path_length_objective(self):
```
patrick-5546, line 33: Naming convention: underscore before helper functions that don't override the default OMPL behavior

local_pathfinding/objectives.py
```diff
@@ -0,0 +1,101 @@
...
+class WindObjective(Objective):
+    def __init__(self, space_information, wind_direction_degrees: float):
+        super().__init__(space_information)
+        self.wind_direction_degrees = wind_direction_degrees
+
+    # This objective function punishes the boat for going up/downwind
+    def motionCost(self, s1: ob.SE2StateSpace, s2: ob.SE2StateSpace) -> ob.Cost:
+        return ob.Cost()
+
+    @staticmethod
```
patrick-5546, line 71: Made these static methods of the class since
- They are only used in this class
- They don't use any instance variables

local_pathfinding/objectives.py
```diff
@@ -0,0 +1,101 @@
...
+    def _is_upwind(wind_direction_radians: float, boat_direction_radians: float) -> bool:
+        raise NotImplementedError
+
+    @staticmethod
+    def _is_downwind(wind_direction_radians: float, boat_direction_radians: float) -> bool:
+        raise NotImplementedError
+
+
+def get_sailing_objective(
```
patrick-5546, line 80: Renamed function to reflect what it does

local_pathfinding/objectives.py
```diff
@@ -0,0 +1,101 @@
...
+    space_information, simple_setup, heading_degrees: float, wind_direction_degrees: float
```
patrick-5546, line 81: I'm unsure of the type of `space_information` and `simple_setup`, if you find out you can add it (optional)

local_pathfinding/objectives.py
patrick-5546: Can add allocate_objectives() to select between the different objectives for debugging purposes at a later date

local_pathfinding/ompl_path.py
```diff
@@ -137,13 +141,21 @@ def _init_simple_setup(self) -> og.SimpleSetup:
...
         simple_setup.setStartAndGoalStates(start, goal)
 
+        # Constructs a space information instance for this simple setup
+        space_information = simple_setup.getSpaceInformation()
+
         # set the optimization objective of the simple setup object
         # TODO: implement and add optimization objective here
-        # simple_setup.setOptimizationObjective(objective)
+
+        objective = get_sailing_objective(
```
patrick-5546, line 150: This should not call self.update_objectives, which is only used for updating not initialization

test/test_objectives.py
```diff
@@ -0,0 +1,58 @@
...
+def test_get_latlon_path_length_objective():
+    with pytest.raises(NotImplementedError):
+        raise NotImplementedError
+
+
+def test_minimum_turning_objective():
+    minimum_turning_objective = objectives.MinimumTurningObjective(
+        PATH._simple_setup.getSpaceInformation(), None, 0
+    )
+    assert minimum_turning_objective is not None
```
patrick-5546, line 38: I think this was a missing test

test/test_objectives.py
patrick-5546, line 1: For test that are specific to an objective, can group them in its own class `Test<Objective>`: https://docs.pytest.org/en/7.4.x/getting-started.html#group-multiple-tests-in-a-class

patrick-5546 (approved):
Left a couple comments for you to go through. I'll approve the PR so that you're not blocked from merging anymore.

test/test_objectives.py
FireBoyAJ24: The issue I faced with adding a test objective class is that the functions will not show up in the testing extension.

PR #42, opened 2023-10-26
Created Three Optimization Objectives

### LINKED ISSUE #24: Create multiple tests for the different Distance optimizers

### LINKED ISSUE #25: Create a Heading Objective

### LINKED ISSUE #46: Tests for coord_systems.py
### Purpose

Make sure all functions in [coord_systems.py](https://github.com/UBCSailbot/local_pathfinding/blob/main/local_pathfinding/coord_systems.py) are tested

### Resources

- tests file: https://github.com/UBCSailbot/local_pathfinding/blob/main/test/test_coord_systems.py

FireBoyAJ24:
### Description

- #25 
- #24 

### Resolves  #25

I have added the distance, minimum turning and upwind/downwind optimization objectives. We can use this as a starting point to build upon, test and verify our optimization objectives.

Note: Two test cases were made for each of the test functions. Later we can add more test cases in a larger issue.

### Verification

- [x] Check whether all the distance objective functions are not producing an error.
- [x] Check whether all the minimum turning objective functions are not producing an error.
- [x] Check whether the wind objective function is not producing an error

### Resources

- https://ubcsailbot.atlassian.net/wiki/spaces/prjt22/pages/1860403201/How+Do+We+Calculate+the+Optimization+Objectives

Note: Please ignore #41 and #32 as this pull request replaces it.

### DIFF
--- a/local_pathfinding/coord_systems.py
+++ b/local_pathfinding/coord_systems.py
@@ -6,7 +6,6 @@ from typing import NamedTuple
 from pyproj import Geod
 
 GEODESIC = Geod(ellps="WGS84")
-M_TO_KM = 0.001
 
 
 class LatLon(NamedTuple):
@@ -49,6 +48,10 @@ def meters_to_km(meters: float) -> float:
     return meters / 1000
 
 
+def km_to_meters(km: float) -> float:
+    return km * 1000
+
+
 def latlon_to_xy(reference: LatLon, latlon: LatLon) -> XY:
     """Convert a geographical coordinate to a 2D Cartesian coordinate given a reference point.
 
@@ -63,7 +66,7 @@ def latlon_to_xy(reference: LatLon, latlon: LatLon) -> XY:
         reference.longitude, reference.latitude, latlon.longitude, latlon.latitude
     )
     true_bearing = math.radians(forward_azimuth_deg)
-    distance = distance_m * M_TO_KM
+    distance = meters_to_km(distance_m)
 
     return XY(
         x=distance * math.sin(true_bearing),
@@ -82,7 +85,7 @@ def xy_to_latlon(reference: LatLon, xy: XY) -> LatLon:
         LatLon: The latitude and longitude in degrees.
     """
     true_bearing = math.degrees(math.atan2(xy.x, xy.y))
-    distance = math.hypot(*xy) / M_TO_KM
+    distance = km_to_meters(math.hypot(*xy))
     dest_lon, dest_lat, _ = GEODESIC.fwd(
         reference.longitude, reference.latitude, true_bearing, distance
     )
--- a/local_pathfinding/coord_systems.py
+++ b/local_pathfinding/coord_systems.py
@@ -80,7 +80,7 @@ def xy_to_latlon(reference: LatLon, xy: XY) -> LatLon:
     true_bearing = math.degrees(math.atan2(xy.x, xy.y))
     distance = math.hypot(*xy) / M_TO_KM
     dest_lon, dest_lat, _ = GEODESIC.fwd(
-        reference.longitude, reference.latitude, math.degrees(true_bearing), distance
+        reference.longitude, reference.latitude, true_bearing, distance
     )
 
     return LatLon(latitude=dest_lat, longitude=dest_lon)
--- a/local_pathfinding/coord_systems.py
+++ b/local_pathfinding/coord_systems.py
@@ -65,3 +65,22 @@ def latlon_to_xy(reference: LatLon, latlon: LatLon) -> XY:
         x=distance * math.sin(true_bearing),
         y=distance * math.cos(true_bearing),
     )
+
+
+def xy_to_latlon(reference: LatLon, xy: XY) -> LatLon:
+    """Convert a 2D Cartesian coordinate to a geographical coordinate given a reference point.
+
+    Args:
+        reference (LatLon): Coordinate that is the origin of the Cartesian coordinate system.
+        xy (XY): Coordinate to be converted to the geographical coordinate system.
+
+    Returns:
+        LatLon: The latitude and longitude in degrees.
+    """
+    true_bearing = math.degrees(math.atan2(xy.x, xy.y))
+    distance = math.hypot(*xy) / M_TO_KM
+    dest_lon, dest_lat, _ = GEODESIC.fwd(
+        reference.longitude, reference.latitude, math.degrees(true_bearing), distance
+    )
+
+    return LatLon(latitude=dest_lat, longitude=dest_lon)
--- a/local_pathfinding/objectives.py
+++ b/local_pathfinding/objectives.py
@@ -195,7 +195,6 @@ class MinimumTurningObjective(Objective):
         Returns:
             float: the turning angle from s2 to s1 (degrees)
         """
-        large_turn_threshold = math.pi / 2
 
         # Calculate the angle of the s1-s2 line segment from North
         path_direction = math.atan2(s2.getX() - s1.getX(), s2.getY() - s1.getY())
@@ -208,12 +207,9 @@ class MinimumTurningObjective(Objective):
 
         # Correct the angle in between [0, pi]
         if turn_size_bias > math.pi:
-            turn_size_unbias = turn_size_bias - 2 * math.pi
+            turn_size_unbias = 2 * math.pi - turn_size_bias
         else:
             turn_size_unbias = turn_size_bias
-
-        if turn_size_unbias > large_turn_threshold:
-            return math.degrees(math.fabs(turn_size_unbias))
 
         return math.degrees(math.fabs(turn_size_unbias))
 
@@ -227,7 +223,6 @@ class MinimumTurningObjective(Objective):
         Returns:
             float: the turning angle from s2 to s1 (degrees)
         """
-        large_turn_threshold = math.pi / 2
 
         # Calculate the angle from s1-goal line segment from North
         global_goal_direction = math.atan2(self.goal_x - s1.getX(), self.goal_y - s1.getY())
@@ -237,12 +232,9 @@ class MinimumTurningObjective(Objective):
 
         # Correct the angle in between [0, pi]
         if turn_size_bias > math.pi:
-            turn_size_unbias = turn_size_bias - 2 * math.pi
+            turn_size_unbias = 2 * math.pi - turn_size_bias
         else:
             turn_size_unbias = turn_size_bias
-
-        if turn_size_unbias > large_turn_threshold:
-            return math.degrees(math.fabs(turn_size_unbias))
 
         return math.degrees(math.fabs(turn_size_unbias))
 
@@ -256,7 +248,6 @@ class MinimumTurningObjective(Objective):
         Returns:
             float: The minimum turning angle between s1-s2 and heading  (degrees)
         """
-        large_turn_threshold = math.pi / 2
 
         # Calculate the angle of the s1-s2 line segment from North
         path_direction = math.atan2(s2.getX() - s1.getX(), s2.getY() - s1.getY())
@@ -266,12 +257,9 @@ class MinimumTurningObjective(Objective):
 
         # Correct the angle in between [0, pi]
         if turn_size_bias > math.pi:
-            turn_size_unbias = turn_size_bias - 2 * math.pi
+            turn_size_unbias = 2 * math.pi - turn_size_bias
         else:
             turn_size_unbias = turn_size_bias
-
-        if turn_size_bias > large_turn_threshold:
-            return math.degrees(math.fabs(turn_size_unbias))
 
         return math.degrees(math.fabs(turn_size_unbias))
 
--- a/local_pathfinding/objectives.py
+++ b/local_pathfinding/objectives.py
@@ -145,7 +145,7 @@ class MinimumTurningObjective(Objective):
     Attributes:
         goal_x (float): The x coordinate of the goal state
         goal_y (float): The y coordinate of the goal state
-        heading (float): The heading of the sailbot in radians [-pi, pi]
+        heading (float): The heading of the sailbot in radians [0, 2pi)
         implementation (str): The implementation of the minimum turning objective function
     """
 
--- a/local_pathfinding/objectives.py
+++ b/local_pathfinding/objectives.py
@@ -158,6 +158,7 @@ class MinimumTurningObjective(Objective):
         self.goal = cs.XY(
             simple_setup.getGoal().getState().getX(), simple_setup.getGoal().getState().getY()
         )
+        assert -180 < heading_degrees <= 180
         self.heading = math.radians(heading_degrees)
         self.method = method
 
@@ -268,12 +269,13 @@ class WindObjective(Objective):
     """Generates a wind objective function
 
     Attributes:
-        wind_direction (float): The direction of the wind in radians [-pi, pi)
+        wind_direction (float): The direction of the wind in radians (-pi, pi]
     """
 
-    def __init__(self, space_information, wind_direction: float):
+    def __init__(self, space_information, wind_direction_degrees: float):
         super().__init__(space_information)
-        self.wind_direction = math.radians(wind_direction)
+        assert -180 < wind_direction_degrees <= 180
+        self.wind_direction = math.radians(wind_direction_degrees)
 
     # This objective function punishes the boat for going up/downwind
     def motionCost(self, s1: ob.SE2StateSpace, s2: ob.SE2StateSpace) -> ob.Cost:
@@ -285,25 +287,29 @@ class WindObjective(Objective):
             s2 (SE2StateInternal): The ending point of the local goal state
 
         Returns:
-            float: The cost of going upwind or downwind
+            ob.Cost: The cost of going upwind or downwind
         """
-        distance = ((s2.getY() - s1.getY()) ** 2 + (s2.getX() - s1.getX()) ** 2) ** 0.5
-        boat_direction_radians = math.atan2(s2.getX() - s1.getX(), s2.getY() - s1.getY())
+        s1_x, s1_y = s1.getX(), s1.getY()
+        s2_x, s2_y = s2.getX(), s2.getY()
+        distance = math.hypot(s2_y - s1_y, s2_x - s1_x)
+        boat_direction_radians = math.atan2(s2_x - s1_x, s2_y - s1_y)
+        assert -math.pi <= boat_direction_radians <= math.pi
 
         if WindObjective.is_upwind(self.wind_direction, boat_direction_radians):
-            return UPWIND_MULTIPLIER * distance
+            cost = UPWIND_MULTIPLIER * distance
         elif WindObjective.is_downwind(self.wind_direction, boat_direction_radians):
-            return DOWNWIND_MULTIPLIER * distance
+            cost = DOWNWIND_MULTIPLIER * distance
         else:
-            return 0.0
+            cost = 0.0
+        return ob.Cost(cost)
 
     @staticmethod
     def is_upwind(wind_direction: float, boat_direction: float) -> bool:
         """Determines whether the boat is upwind or not and its associated cost
 
         Args:
-            wind_direction (float): The true wind direction (radians). [-pi, pi)
-            boat_direction (float): The direction of the boat (radians). [-pi, pi)
+            wind_direction (float): The true wind direction (radians). (-pi, pi]
+            boat_direction (float): The direction of the boat (radians). [-pi, pi]
 
         Returns:
             bool: The cost associated with the upwind direction
@@ -318,8 +324,8 @@ class WindObjective(Objective):
         """Generates the cost associated with the downwind direction
 
         Args:
-            wind_direction (float): The true wind direction (radians). [-pi, pi)
-            boat_direction_radians (float)): The direction of the boat (radians). [-pi, pi)
+            wind_direction (float): The true wind direction (radians). (-pi, pi]
+            boat_direction_radians (float)): The direction of the boat (radians). [-pi, pi]
 
         Returns:
             bool: The cost associated with the downwind direction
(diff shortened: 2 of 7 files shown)


patrick-5546:
> Note: Test cases have not been made as requested in https://github.com/UBCSailbot/local_pathfinding/issues/24 since that is a separate major issue to tackle.

If this is the case does this PR resolve the issue, or should we leave it open even after merging in this PR?

local_pathfinding/objectives.py
```diff
@@ -20,61 +29,322 @@ def __init__(self, space_information):
         self.space_information = space_information
 
     def motionCost(self, s1: ob.SE2StateSpace, s2: ob.SE2StateSpace) -> ob.Cost:
-        raise NotImplementedError
+        return ob.Cost(0.0)
```
patrick-5546: Don't we want to raise a `NontImplementedError` if an objective doesn't implement `motionCost`?

local_pathfinding/objectives.py
```diff
@@ -20,61 +29,322 @@ def __init__(self, space_information):
...
+        space_information (class): The space information of the OMPL problem
+    """
+
     def __init__(self, space_information):
         super().__init__(space_information)
+        self.space_information = space_information
 
-    def motionCost(self, s1: ob.SE2StateSpace, s2: ob.SE2StateSpace) -> ob.Cost:
-        return ob.Cost()
+    def motionCost(self, s1, s2):
```
patrick-5546, line 46: Did not copy over type hints

local_pathfinding/objectives.py
```diff
@@ -20,61 +29,322 @@ def __init__(self, space_information):
...
+            * 6378
+        )
 
 
 class MinimumTurningObjective(Objective):
-    def __init__(self, space_information, simple_setup, heading_degrees: float):
+    """Generates a minimum turning objective function
+
+    Attributes:
+        space_information (class): The space information of the OMPL problem
```
patrick-5546: Move attribute documentation to `Objective`, where it is set

local_pathfinding/objectives.py
```diff
@@ -20,61 +29,322 @@ def __init__(self, space_information):
...
+        goal_x (float): The x coordinate of the goal state
+        goal_y (float): The y coordinate of the goal state
+        heading_degrees (float): The heading of the sailbot in degrees
+    """
+
+    def __init__(self, space_information, simple_setup, heading_degrees):
         super().__init__(space_information)
-        self.simple_setup = simple_setup
-        self.heading_degrees = heading_degrees
+        self.space_information = space_information
```
patrick-5546: Redundant, as it is set in `Objective`

test/test_objectives.py
```diff
@@ -17,42 +29,239 @@ def test_distance_objective():
...
+            math.radians(PATH.state.windDirection), math.radians(PATH.state.headingDirection)
+        )
+        == expected
+    )
+
+
+""" Tests for bound_to_180() """
+
 
+def test_bound_basic1():
```
patrick-5546: Parameterize

test/test_objectives.py
```diff
@@ -17,42 +29,239 @@ def test_distance_objective():
...
-    with pytest.raises(NotImplementedError):
-        raise NotImplementedError
 
+""" Tests for is_angle_between() """
 
-def test_headingPathTurnCost():
-    with pytest.raises(NotImplementedError):
-        raise NotImplementedError
 
+def test_between_basic1():
```
patrick-5546: Parameterize

patrick-5546 (commented):
General comments: look at the files changed tab to see the latest changes from the skeleton code PR that are in main that were not applied to this branch
- Add type hints to all functions except for `simple_setup` and `space_information` arguments
- Function naming convention: only OMPL functions should be camelCase, everything else should be snake_case (no need to start with `_`)

FireBoyAJ24:
> > Note: Test cases have not been made as requested in #24 since that is a separate major issue to tackle.
> 
> If this is the case does this PR resolve the issue, or should we leave it open even after merging in this PR?

The PR resolves the basic  testing of each function. I want to have a larger issue that requests the addition of more test cases.

local_pathfinding/objectives.py
```diff
@@ -24,57 +35,316 @@ def motionCost(self, s1: ob.SE2StateSpace, s2: ob.SE2StateSpace) -> ob.Cost:
...
+                       (currently it is returning a object)
+        """
 
-    def _get_latlon_path_length_objective(self, s1, s2):
-        raise NotImplementedError
+        # Generates the euclidean distance between two points
+        # euclideanPathObjective = self.get_euclidean_path_length_objective(s1, s2)
+
+        # Generates the latlon distance between two points
+        latlonPathObjective = self.get_latlon_path_length_objective(s1, s2)
```
patrick-5546: Could add an attribute `implementation` to select between the different implementations

local_pathfinding/objectives.py
```diff
@@ -24,57 +35,316 @@ def motionCost(self, s1: ob.SE2StateSpace, s2: ob.SE2StateSpace) -> ob.Cost:
...
+
+        Args:
+            wind_direction_radians (float): The true wind direction (radians). [-pi, pi)
+            boat_direction_radians (float)): The direction of the boat (radians). [-pi, pi)
+
+        Returns:
+            bool: The cost associated with the downwind direction
+        """
+        downwind_wind_direction_radians = math.radians(
+            WindObjective.bound_to_180(wind_direction_radians + math.pi)
```
patrick-5546: Different units, bounding radians to 180 degrees

local_pathfinding/objectives.py
```diff
@@ -24,57 +35,316 @@ def motionCost(self, s1: ob.SE2StateSpace, s2: ob.SE2StateSpace) -> ob.Cost:
...
+
+        if turn_size_bias > large_turn_threshold:
+            return math.degrees(math.fabs(turn_size_unbias))
+
+        return math.degrees(math.fabs(turn_size_unbias))
 
 
 class WindObjective(Objective):
-    def __init__(self, space_information, wind_direction_degrees: float):
+    """Generates a wind objective function"""
```
patrick-5546: Document attribute `wind_direction` We don't need to include units in variable name, but it should be stated in docstrings.

patrick-5546: Also I think if you make this radians you don't have to convert anywhere else in `WindObjective`

patrick-5546 (commented):
I fixed the rest of the inconsistencies with the main branch. In the future, you can avoid this by double checking the files changed tab.

One other note is that we need to be able to update the objective attributes. For example, `wind_direction` or `goal_x`, `goal_y` will change during runtime.

And just a reminder to pull this branch to get my changes before making new changes to avoid merge conflicts.

local_pathfinding/objectives.py
```diff
@@ -37,8 +37,9 @@ def motionCost(self, s1: ob.SE2StateSpace, s2: ob.SE2StateSpace) -> ob.Cost:
 class DistanceObjective(Objective):
     """Generates a distance objective function"""
```
patrick-5546: Document `implementation` attribute for both distance and minimum turning objectives

local_pathfinding/objectives.py
```diff
@@ -51,14 +52,12 @@ def motionCost(self, s1: ob.SE2StateSpace, s2: ob.SE2StateSpace) -> ob.Cost:
...
-        # Generates the latlon distance between two points
-        latlonPathObjective = self.get_latlon_path_length_objective(s1, s2)
-
-        return latlonPathObjective
+        if self.implementation == "euclidean":
+            # Generates the euclidean distance between two points
+            return self.get_euclidean_path_length_objective(s1, s2)
+        elif self.implementation == "latlon":
+            # Generates the latlon distance between two points
+            return self.get_latlon_path_length_objective(s1, s2)
```
patrick-5546: else raise `ValueError(f"Implementation {self.implementation} not supported")` for both distance and minimum turning objectives

local_pathfinding/objectives.py
```diff
@@ -37,8 +37,9 @@ def motionCost(self, s1: ob.SE2StateSpace, s2: ob.SE2StateSpace) -> ob.Cost:
 class DistanceObjective(Objective):
     """Generates a distance objective function"""
 
-    def __init__(self, space_information):
+    def __init__(self, space_information, implementation="euclidean"):
```
patrick-5546: Best way would be to create an enum type for DistanceMethod and TurnCost

For example
```py
from enum import Enum, auto

class Color(Enum):
    RED = auto()
    BLUE = auto()
    GREEN = auto()
```

Resource: https://docs.python.org/3/howto/enum.html

patrick-5546: Enums can go at the top of this file

local_pathfinding/objectives.py
```diff
@@ -111,14 +110,17 @@ class MinimumTurningObjective(Objective):
     Attributes:
         goal_x (float): The x coordinate of the goal state
         goal_y (float): The y coordinate of the goal state
-        heading_degrees (float): The heading of the sailbot in degrees
+        heading (float): The heading of the sailbot in radians
```
patrick-5546: Range is [0, 2pi)

https://github.com/UBCSailbot/custom_interfaces/blob/main/msg/HelperHeading.msg

local_pathfinding/objectives.py
```diff
@@ -233,11 +235,15 @@ def heading_path_turn_cost(self, s1: ob.SE2StateSpace, s2: ob.SE2StateSpace):
 
 
 class WindObjective(Objective):
-    """Generates a wind objective function"""
+    """Generates a wind objective function
 
-    def __init__(self, space_information, wind_direction_degrees):
+    Attributes:
+        wind_direction (float): The direction of the wind in radians
```
patrick-5546: Range is [-pi, pi)

https://github.com/UBCSailbot/custom_interfaces/blob/main/msg/WindSensor.msg

local_pathfinding/objectives.py
```diff
@@ -24,66 +35,308 @@ def motionCost(self, s1: ob.SE2StateSpace, s2: ob.SE2StateSpace) -> ob.Cost:
...
+        Returns:
+            float: The minimum turning angle between s1-s2 and heading  (degrees)
+        """
+        large_turn_threshold = math.pi / 2
+
+        # Calculate the angle of the s1-s2 line segment from North
+        path_direction = math.atan2(s2.getX() - s1.getX(), s2.getY() - s1.getY())
+
+        # Calculate turn size
+        turn_size_bias = math.fabs(path_direction - self.heading)
```
patrick-5546: Have you thought about the bounds of each variable? You may have to account for crossing the 2pi/0 line. And wouldn't you want to get the smaller of the angles between them?

local_pathfinding/objectives.py
```diff
@@ -1,7 +1,17 @@
...
+import math
+
 from ompl import base as ob
 
+# Upwind downwind cost multipliers
+UPWIND_MULTIPLIER = 3000.0
+DOWNWIND_MULTIPLIER = 3000.0
+
+# Upwind downwind constants
+UPWIND_MAX_ANGLE_DEGREES = 40.0
```
patrick-5546: Can change to `UPWIND_MAX_ANGLE_RADIANS = math.radians(40.0)`

local_pathfinding/objectives.py
```diff
@@ -24,66 +35,308 @@ def motionCost(self, s1: ob.SE2StateSpace, s2: ob.SE2StateSpace) -> ob.Cost:
...
     space_information, simple_setup, heading_degrees: float, wind_direction_degrees: float
 ) -> ob.OptimizationObjective:
     objective = ob.MultiOptimizationObjective(si=space_information)
-    objective.addObjective(objective=DistanceObjective(space_information), weight=1.0)
+    objective.addObjective(objective=DistanceObjective(space_information, "latlon"), weight=1.0)
     objective.addObjective(
-        objective=MinimumTurningObjective(space_information, simple_setup, heading_degrees),
+        objective=MinimumTurningObjective(
+            space_information, simple_setup, heading_degrees, "goal_heading"
+        ),
```
patrick-5546: Any reason why they override the default `implementation`s?

patrick-5546 (commented):
Next round of comments

local_pathfinding/objectives.py
FireBoyAJ24: Just because.....

local_pathfinding/objectives.py
FireBoyAJ24: When I use `math.atan2(x, y)` then my angles gets bounded between (-pi, pi]. (Note: I am saying `math.atan2(x, y)` and not `math.atan2(y, x)`)

local_pathfinding/coord_systems.py
```diff
@@ -65,3 +65,22 @@ def latlon_to_xy(reference: LatLon, latlon: LatLon) -> XY:
         x=distance * math.sin(true_bearing),
         y=distance * math.cos(true_bearing),
     )
+
+
+def xy_to_latlon(reference: LatLon, xy: XY) -> LatLon:
```
patrick-5546, line 77: @jamenkaye will be writing tests for this

jamenkaye: Added tests!

jamenkaye:
Good thing I added tests for `xy_to_latlon()`, I caught a bug where `true_bearing` was converted to degrees twice :)

@FireBoyAJ24 I noticed `test_get_latlon_path_length_objective()` was failing too. Any idea what's up with this? I looked for a little while but couldn't find the issue.

jamenkaye:
Adding in this information from Jay's entry in the Pathfinding meeting notes:

I was integrating the `xy_to_lat-lon` function (coord_systems.py) into the `_get_latlon_path_length_objective` (objectives.py) function.

Issue: My expected distance between two lat-lon coordinates does not agree with the calculated distance from GEODESIC. In short `test_get_latlon_path_length_objectivefunction` (test_objectives.py) is spitting out a not expected error.

Other than that we need to do another round of review to ensure all the formatting is corrected and all the tests have passed.

local_pathfinding/objectives.py
```diff
@@ -1,7 +1,35 @@
...
+
 from ompl import base as ob
 
+# Upwind downwind cost multipliers
+UPWIND_MULTIPLIER = 3000.0
+DOWNWIND_MULTIPLIER = 3000.0
+
+# Upwind downwind constants
+UPWIND_MAX_ANGLE_RADIANS = math.radians(40.0)
+DOWNWIND_MAX_ANGLE_RADIANS = math.radians(20.0)
```
jamenkaye: Really these are minimum angles, right? 
We want to communicate that we can't sail closer to irons than 40 degrees. Maybe `HIGHEST_UPWIND_ANGLE` and `LOWEST_DOWNWIND_ANGLE` would be more descriptive

local_pathfinding/objectives.py
```diff
@@ -24,66 +53,313 @@ def motionCost(self, s1: ob.SE2StateSpace, s2: ob.SE2StateSpace) -> ob.Cost:
...
+
+        Args:
+            s1 (SE2StateInternal): The starting point of the local start state
+            s2 (SE2StateInternal): The ending point of the local goal state
+
+        Returns:
+            ob.Cost: the turning angle from s2 to s1 (degrees)
+        """
+
+        # Calculate the angle of the s1-s2 line segment from North
```
jamenkaye: more specifically, the comment could say `# Calculate the true bearing of s2 from s1`

local_pathfinding/objectives.py
```diff
@@ -24,66 +53,313 @@ def motionCost(self, s1: ob.SE2StateSpace, s2: ob.SE2StateSpace) -> ob.Cost:
...
+        path_direction = math.atan2(s2.getX() - s1.getX(), s2.getY() - s1.getY())
+
+        # Calculate the angle from s1-goal line segment from North
```
jamenkaye: similarly, this could say `# Calculate the true bearing of the goal from s1`

local_pathfinding/objectives.py
```diff
@@ -24,66 +53,313 @@ def motionCost(self, s1: ob.SE2StateSpace, s2: ob.SE2StateSpace) -> ob.Cost:
...
+        # Calculate the uncorrected turn size [0, 2pi]
+        turn_size_bias = math.fabs(angle1 - angle2)
+
+        # Correct the angle in between [0, pi]
+        if turn_size_bias > math.pi:
+            turn_size_unbias = 2 * math.pi - turn_size_bias
+        else:
+            turn_size_unbias = turn_size_bias
+
+        return math.degrees(math.fabs(turn_size_unbias))
```
jamenkaye, line 265: I could be wrong but it seems like the `fabs` function isn't needed in this line? In the `if` block, `turn_size_unbias` is $2 \pi - \text{ angle between } (\pi, 2\pi]$, and in the else block, `turn_size_unbias` is already positive due to the `fabs` on line 246.

local_pathfinding/objectives.py
```diff
@@ -24,66 +53,313 @@ def motionCost(self, s1: ob.SE2StateSpace, s2: ob.SE2StateSpace) -> ob.Cost:
...
+
+        Args:
+            s1 (SE2StateInternal): The starting point of the local start state
+            s2 (SE2StateInternal): The ending point of the local goal state
+
+        Returns:
+            ob.Cost: the turning angle from s2 to s1 (degrees)
+        """
+
+        # Calculate the angle from s1-goal line segment from North
```
jamenkaye: how about `# Calculate the true bearing of the goal from s1` ?

local_pathfinding/objectives.py
```diff
@@ -24,66 +53,313 @@ def motionCost(self, s1: ob.SE2StateSpace, s2: ob.SE2StateSpace) -> ob.Cost:
...
+
+        Args:
+            s1 (SE2StateInternal): The starting point of the local start state
+            s2 (SE2StateInternal): The ending point of the local goal state
+
+        Returns:
+            ob.Cost: The minimum turning angle between s1-s2 and heading  (degrees)
+        """
+
+        # Calculate the angle of the s1-s2 line segment from North
```
jamenkaye: same thing, but how about `# Calculate the true bearing of s2 from s1`?

local_pathfinding/objectives.py
```diff
@@ -24,66 +53,313 @@ def motionCost(self, s1: ob.SE2StateSpace, s2: ob.SE2StateSpace) -> ob.Cost:
...
+           relation to the wind
+
+        Args:
+            s1 (SE2StateInternal): The starting point of the local start state
+            s2 (SE2StateInternal): The ending point of the local goal state
+
+        Returns:
+            float: The cost of going upwind or downwind
+        """
+        distance = ((s2.getY() - s1.getY()) ** 2 + (s2.getX() - s1.getX()) ** 2) ** 0.5
```
jamenkaye: I saw you were using a hypotenuse function elsewhere, should that maybe be used here?

local_pathfinding/objectives.py
```diff
@@ -24,66 +53,313 @@ def motionCost(self, s1: ob.SE2StateSpace, s2: ob.SE2StateSpace) -> ob.Cost:
...
+        downwind_wind_direction = (wind_direction + math.pi) % (2 * math.pi)
+
+        theta_min = downwind_wind_direction - DOWNWIND_MAX_ANGLE_RADIANS
+
+        theta_max = downwind_wind_direction + DOWNWIND_MAX_ANGLE_RADIANS
+
+        return WindObjective.is_angle_between(theta_min, boat_direction, theta_max)
+
+    @staticmethod
+    def is_angle_between(first_angle: float, middle_angle: float, second_angle: float) -> bool:
```
jamenkaye, line 354: for Jamen: Does this work for unbounded angles? the `is_upwind` function could give unbounded angles.

jamenkaye (commented):
Reviewed!

local_pathfinding/objectives.py
jamenkaye: Made `is_angle_between()` handle un bounded angles, because of functions like `is_upwind()`. In `is_upwind()`, we could have `wind_angle` = pi (or nearly pi), then when we add on the `HIGHEST_UPWIND_ANGLE_RADIANS` we could pass an unbounded angle to `is_angle_between()`.

local_pathfinding/objectives.py
jamenkaye: This works though, its fine 👍

local_pathfinding/objectives.py
jamenkaye: Since an extra fabs() function is such a small inefficiency, but taking it out could possibly risk breaking something, I'll just leave it in.

jamenkaye (approved):
Looks good!

I think there might be an unecessary `fabs()` function on line 256: `return math.degrees(math.fabs(turn_size_unbias))`, but this is a very small inefficiency and I don't want to risk breaking anything haha

jamenkaye:
... But I'd still love a second opinion :) @patrick-5546

patrick-5546:
Summarizing the changes I made
- Fix types and bounds
- Refactor to reduce functions/tests that depend on OMPL

PR #43, opened 2023-10-26
Make jamenkaye codeowner

patrick-5546:
@jamenkaye adding you as codeowner so that you get notified when PRs are made. My thinking is that my reviews will more focus more on code quality and architecture, whereas yours can focus more on functionality and implementation


jamenkaye:
Thanks! I'll make sure to focus on functionality and implementation when doing code reviews 👍

PR #61, opened 2023-11-14
Mock global path implementation

### LINKED ISSUE #49: Mock global path
### Purpose

Run our software without having access to the actual global path

### Description

- Subscribe to `gps` to get current position
- Publish to `global_path`
- Constant in file that defines destination latlon
- Interpolate points between current position and destination and publish to global path

### Resources

- External interface topic names and types: https://github.com/UBCSailbot/custom_interfaces/blob/main/README.md#project-wide-external-interfaces
- Example node: https://github.com/UBCSailbot/local_pathfinding/blob/main/local_pathfinding/node_navigate.py
- ROS tutorial on nodes: https://docs.ros.org/en/humble/Tutorials/Beginner-CLI-Tools/Understanding-ROS2-Nodes/Understanding-ROS2-Nodes.html

SPDonaghy:
### Description
- Resolves #49 

Mock Global Path Node:

- subscribes to GPS
- accepts a global path parameter (List) which can be an actual path (list of waypoints) or a single element list (Destination)
- publishes to global_path

Other Features:

- Path interpolation function: This function analyzes the given path, including the starting position, and fills in any gaps (where waypoints are too far apart) with a subpath, using the same generate_path function
- interval spacing function: a helper function for the interpolator, which returns an array of all the actual interval lengths for a given path
- 3D plotting script
- Purge script for deleting any generated path files in the global_paths directory
- Instead of a service, in order to force the path to update, from the command line, you can just set the parameter `force` to true. The checks in the callback will be bypassed, the path updated, and then the param is set to false again.

Path Builder GUI Features:

- Export to CSV
- Clear
- 3D Plot (which exports to a CSV in the process)

would like to add:

- import paths from csv files
- edit paths 
- join paths
- travel time estimate
- LatLon text fields for more precise point selection
- maybe a stylesheet and logo

### Verification
No verification at this time

### Resources
-  [Node file used as an example](https://github.com/UBCSailbot/local_pathfinding/blob/main/local_pathfinding/node_navigate.py)


patrick-5546 (commented):
@SPDonaghy updated the ROS infrastructure so that `ros2 launch local_pathfinding main_launch.py` launches your node. If you want to make any changes to the files outside of `local_pathfinding/local_pathfinding`, you will have to build local_pathfinding to apply them.

SPDonaghy:
> @SPDonaghy updated the ROS infrastructure so that `ros2 launch local_pathfinding main_launch.py` launches your node. If you want to make any changes to the files outside of `local_pathfinding/local_pathfinding`, you will have to build local_pathfinding to apply them.

Okay, I made some changes in the latest commit to ensure that the mgp node runs upon launch. Seems to work as expected.

patrick-5546:
@hhenry01 any update on the remote transceiver endpoint?

hhenry01:
> @hhenry01 any update on the remote transceiver endpoint?

Use http://localhost:8081/global-path

patrick-5546:
What should the format be? JSON?

launch/main_launch.py
```diff
@@ -65,8 +65,11 @@ def setup_launch(context: LaunchContext) -> List[Node]:
     Returns:
         List[Node]: Nodes to launch.
     """
-    launch_description_entities = list()
+    mode = LaunchConfiguration("mode").perform(context)
```
patrick-5546, line 68: I'm thinking we should create a new boolean launch argument `mock_global_path`. To determine whether it is run. NET would use this argument to determine whether to publish global path as well

launch/main_launch.py
```diff
@@ -65,8 +65,11 @@ def setup_launch(context: LaunchContext) -> List[Node]:
...
+    launch_description_entities = []
     launch_description_entities.append(get_navigate_node_description(context))
+    if mode == "development":
```
patrick-5546, line 71: I'm thinking we should create a new boolean launch argument `mock_global_path`. To determine whether it is run. NET would use this argument to determine whether to publish global path as well. So only run node if in development mode and `mock_global_path` is true

local_pathfinding/node_mock_global_path.py
```diff
@@ -0,0 +1,222 @@
...
+import rclpy
+from custom_interfaces.msg import GPS, HelperLatLon, Path
+from custom_interfaces.srv import GlobalPath
+from rclpy.node import Node
+
+# Destination is hardcoded temporarily as a single element list
+MOCK_DESTINATION = []
+# List of doubles is an allowed type for a ros2 parameter
+MOCK_DESTINATION.append("49.263,123.138")
+# Mock gps data to get things running until we have a running gps node
```
patrick-5546: Why not do nothing until received GPS msg?

local_pathfinding/node_mock_global_path.py
```diff
@@ -0,0 +1,222 @@
...
+MOCK_GPS = GPS(lat_lon=HelperLatLon(latitude=48.9594, longitude=-123.3634))
+
+NUM_INTERVALS = 100
+GLOBAL_PATH_UPDATE_DELAY = 10
```
patrick-5546: These as well as `MOCK_DESTINATION` can be ROS parameters where the default value is set in https://github.com/UBCSailbot/sailbot_workspace/tree/main/src/global_launch/config

local_pathfinding/node_mock_global_path.py
```diff
@@ -0,0 +1,222 @@
...
+        gps_sub (Subscription): Subscribe to a `GPS` msg.
+
+    Services:
+        global_path_srv (Service): Service to send the global path to node_navigate,
+        as a request message.
+
+    Attributes from subscribers:
+        gps (GPS): Data from the GPS sensor.
+
+    Attributes:
```
patrick-5546, line 226: All attributes should be documented here and initialized in `__init__()`, even if it is to `None`. For example, `future`

local_pathfinding/node_mock_global_path.py
```diff
@@ -0,0 +1,222 @@
...
+            [
+                rclpy.parameter.Parameter(
+                    "global_path_param", rclpy.Parameter.Type.STRING_ARRAY, MOCK_DESTINATION
+                )
+            ]
+        )
+        # services
+        self.client = self.create_client(GlobalPath, "global_path_srv")
+
+        while not self.client.wait_for_service(timeout_sec=1.0):
```
patrick-5546: What does this check? For node_navigate to be running?

local_pathfinding/node_mock_global_path.py
```diff
@@ -0,0 +1,222 @@
...
+            self.get_logger().info("connection to node_navigate failed, waiting again...")
```
patrick-5546: Use `warn()` instead of `info()`

local_pathfinding/node_mock_global_path.py
```diff
@@ -0,0 +1,222 @@
...
+
+        self.set_parameters(
+            [
+                rclpy.parameter.Parameter(
+                    "global_path_param", rclpy.Parameter.Type.STRING_ARRAY, MOCK_DESTINATION
+                )
+            ]
+        )
+        # services
+        self.client = self.create_client(GlobalPath, "global_path_srv")
```
patrick-5546: Use key word arguments (include argument names). In my mind this is good to do unless the argument and variable names are the same

local_pathfinding/node_mock_global_path.py
```diff
@@ -0,0 +1,222 @@
...
+        self.get_logger().info(f"Received data from {self.gps_sub.topic}: {msg}")
+        self.gps = msg
+
+    # Service callbacks
+    def global_path_callback(self):
+        """Check if the global path parameter has changed, on a regular time interval.
+        If it has changed, the new path is sent to node_navigate
+
+        Global path can be changed through a ros2 param set command in the terminal
+        or by modifying mock_global_path.csv
```
patrick-5546: How I imagine this node could be used:
1. Update destination -> new global path with evenly spaced waypoints to destination
2. Update path -> append path to new global path with evenly spaced waypoints to first waypoint in path

But really 1 is just a special case of 2.

patrick-5546: I like being able to modify the csv or setting the parameter. What would also be nice is to specify a path to a different CSV

local_pathfinding/node_mock_global_path.py
```diff
@@ -0,0 +1,222 @@
...
+                reader = csv.reader(file)
+                reader.__next__()
+                global_path = []
+                for row in reader:
+                    global_path.append(row[0] + "," + row[1])
+
+            self.path_mod_tmstmp = path_mod_tmstmp
+
+            # Set the new global path parameter
+            self.set_parameters(
```
patrick-5546: This would override it being set in CLI right? Do we want CSV change or CLI update to dominate? Should probably be documented somewhere

local_pathfinding/node_mock_global_path.py
```diff
@@ -0,0 +1,222 @@
...
+            return
+
+        self.global_path = global_path
+
+    def send_global_path(self):
+        self.req.global_path = MockGlobalPath.str_list_to_path(self.global_path)
+        self.future = self.client.call_async(self.req)
+
+    def _all_subs_active(self) -> bool:
+        return True  # TODO: this line is a placeholder, delete when mocks can be run
```
patrick-5546: `return self.gps is not None`

local_pathfinding/node_mock_global_path.py
```diff
@@ -0,0 +1,222 @@
...
+
+    def _log_inactive_subs_warning(self):
+        # TODO: log which subscribers are inactive
+        self.get_logger().warning("There are inactive subscribers")
```
patrick-5546: `"Waiting for GPS to be published"`

local_pathfinding/node_mock_global_path.py
```diff
@@ -0,0 +1,222 @@
...
+        intermediate points"""
+
+        if not self._all_subs_active():
+            self._log_inactive_subs_warning()
+            return
+
+        global_path = self.get_parameter("global_path_param")._value
+
+        # Check if global path parameter is just a destination point
+        if len(global_path) < 2:
```
patrick-5546: Should be in its own function so that you can test it

local_pathfinding/node_mock_global_path.py
```diff
@@ -0,0 +1,222 @@
...
+    rclpy.spin(node=mock_global_path)
+
+    mock_global_path.destroy_node()
+    rclpy.shutdown()
+
+
+class MockGlobalPath(Node):
+    """Stores and sends the global path to the navigate node.
+
+    Subscribers:
```
patrick-5546, line 215: Maybe we can be more specific in `node_*.py`: topic subscribers/publishers, service clients/servers

patrick-5546 (commented):
Great start

launch/main_launch.py
patrick-5546: New plan is to create a parameter, will do that in part 2

launch/main_launch.py
patrick-5546: New plan is to create a parameter, will do that in part 2

local_pathfinding/node_mock_global_path.py
SPDonaghy: add comment to remove this once NET has published gps

local_pathfinding/node_mock_global_path.py
patrick-5546: - Derive `self.num_intervals = distance(destination, current_position) / distance_between_waypoints=30km`
- Make `distance_between_waypoints` the parameter

patrick-5546: Can use geodesic function where you give starting point, distance, and course to get next waypoint (see latlon2xy)

local_pathfinding/node_mock_global_path.py
patrick-5546: Change parameter from global path to file path to global path csv

resource/mock_global_path.csv
patrick-5546, line 1: Put in root directory `global_paths/`

local_pathfinding/node_mock_global_path.py
```diff
@@ -0,0 +1,232 @@
...
+from rclpy.node import Node
+
+ELLIPSOID = Geod(ellps="WGS84")
+M_to_KM = 0.001
+
+# Mock gps data to get things running until we have a running gps node
+# TODO Remove when NET publishes GPS
+MOCK_GPS = GPS(lat_lon=HelperLatLon(latitude=48.9594, longitude=-123.3634))
+
+DEFAULT_PATH = "local_pathfinding/global_paths/mock_global_path.csv"
```
patrick-5546: Should be absolute path so that you can run from any directory

local_pathfinding/node_mock_global_path.py
```diff
@@ -0,0 +1,232 @@
...
+INTERVAL_SPACING = 30.0  # km
+GLOBAL_PATH_UPDATE_DELAY = 5.0  # seconds
```
patrick-5546: Technically this is the period of running the callback right, not updating the path

patrick-5546: Any reason not to just use the global parameter `pub_period_sec`?

local_pathfinding/node_mock_global_path.py
```diff
@@ -0,0 +1,232 @@
...
+                    type_=rclpy.Parameter.Type.STRING,
+                    value=DEFAULT_PATH,
+                ),
+                rclpy.parameter.Parameter(
+                    name="interval_spacing",
+                    type_=rclpy.Parameter.Type.DOUBLE,
+                    value=INTERVAL_SPACING,
+                ),
+            ]
+        )
```
patrick-5546: Can you make a PR to add these to the global params file in sailbot workspace instead
https://github.com/UBCSailbot/sailbot_workspace/tree/main/src/global_launch/config

local_pathfinding/node_mock_global_path.py
```diff
@@ -0,0 +1,232 @@
...
+        Global path can be changed by modifying mock_global_path.csv or setting the
+        global_path_filepath parameter to a filepath to a new csv file.
+        """
+
+        file_path = self.get_parameter("global_path_filepath")._value
+
+        # check when global path was changed last
+        path_mod_tmstmp = time.ctime(os.path.getmtime(file_path))
+
+        if path_mod_tmstmp != self.path_mod_tmstmp:
```
patrick-5546: What if the file path has changed and the new file is older than the original file?

local_pathfinding/node_mock_global_path.py
```diff
@@ -0,0 +1,232 @@
...
+                # skip header
+                reader.__next__()
+                for row in reader:
+                    global_path.waypoints.append(
+                        HelperLatLon(latitude=float(row[0]), longitude=float(row[1]))
+                    )
+
+            self.path_mod_tmstmp = path_mod_tmstmp
+
+            if self.gps is not None:
```
patrick-5546: Can call `_all_subs_active()` here?

local_pathfinding/node_mock_global_path.py
patrick-5546: Also there is a lot of nesting here, would it be possible to flatten it out? For example, you could probably bring the gps is none check to this condition

local_pathfinding/node_mock_global_path.py
```diff
@@ -0,0 +1,232 @@
...
+                        if (
+                            global_path.waypoints[i].latitude
+                            == global_path.waypoints[i - 1].latitude
+                            and global_path.waypoints[i].longitude
+                            == global_path.waypoints[i - 1].longitude
+                        ):
+                            self.get_logger().warn(
+                                f"Found duplicate consecutive waypoint {global_path.waypoints[i]}"
+                            )
+                            # global_path.waypoints.pop(i)
```
patrick-5546: Why do you check for this? Did you mean to comment out the pop? I ran into this error when using the default global path

local_pathfinding/node_mock_global_path.py
```diff
@@ -0,0 +1,232 @@
...
+
+                    self.global_path = global_path
+
+                msg = self.global_path
+                self.global_path_pub.publish(msg)
+                self.get_logger().info(f"Publishing to {self.global_path_pub.topic}: {msg}")
+
+            else:
+                self.get_logger().warn("No GPS data")
```
patrick-5546: How I was intending this to work is that this warning would be defined in `_log_inactive_subs_warning()`, which will be called by `_all_subs_active()`

local_pathfinding/node_mock_global_path.py
```diff
@@ -0,0 +1,232 @@
...
+
+    def generate_path(self, dest: HelperLatLon):
```
patrick-5546: If you make this static and add `interval_spacing` and `pos` as arguments it will simplify your test so that it doesn't depend on ROS anymore

local_pathfinding/global_paths/mock_global_path.csv
patrick-5546, line 1: Please move `global_paths/` to the root of this repository, and not in `local_pathfinding/` which is just for python source files

local_pathfinding/global_paths/path_2.csv
patrick-5546, line 1: Would it be easy to create a simple script in this directory to visualize/plot a global path? This could be helpful for a sanity check. For example you could call it with `python3 plot_global_path.py mock_global_path.csv`

patrick-5546: Or if you don't have time to get it in by Friday you can create an issue for this

local_pathfinding/node_mock_global_path.py
SPDonaghy: No not really, I just didn't want it to check for changes that frequently. But I can change it

local_pathfinding/node_mock_global_path.py
SPDonaghy: It should still work as long as the time stamps are not identical

local_pathfinding/node_mock_global_path.py
SPDonaghy: Just a thought I had, when I was coming up with tests. But I don't know if its going to really be an issue. I think I will remove it and fix the csv file

local_pathfinding/global_paths/path_2.csv
SPDonaghy: Sure! But yes I likely won't have time by then

local_pathfinding/node_mock_global_path.py
patrick-5546: I think it would be better to be more explicit. You can create an instance variable for file path as well and update it at the top from the parameter

local_pathfinding/global_paths/path_2.csv
patrick-5546: Stretch goal, interactively create a global path file by selecting points on a map. Maybe this could be integrated into the website?

local_pathfinding/node_mock_global_path.py
```diff
@@ -0,0 +1,224 @@
...
+        if not self._all_subs_active():
+            self._log_inactive_subs_warning()
+
+        file_path = self.get_parameter("global_path_filepath")._value
+
+        # check when global path was changed last
+        path_mod_tmstmp = time.ctime(os.path.getmtime(file_path))
+
+        # Only publish if the path has changed
+        if (path_mod_tmstmp != self.path_mod_tmstmp) or (self.file_path != file_path):
```
patrick-5546: log info whether publishing because of timestamp or file path change. To properly implement this you would want to check file path before timestamp

local_pathfinding/global_paths/path_2.csv
SPDonaghy: Sounds good. what are your thoughts on the generate path function overwriting the source csv file if it generates a path?

SPDonaghy: overwriting, with the generated path that is

SPDonaghy:
Test out the visualizer with the command: `python3 local_pathfinding/plot_global_path.py mock_global_path.csv 30.0`

local_pathfinding/global_paths/path_2.csv
SPDonaghy: Test out the visualizer with the command: `python3 local_pathfinding/plot_global_path.py mock_global_path.csv 30.0`

patrick-5546:
I see that you now write the global path to the file path. What do you see as the usecase for this? In my mind the node would only need to read from the files: the parameter could be used to quickly/temporarily tests a new path, and if we want to update or create a new file it makes sense to me for this to be a manual choice/process.

local_pathfinding/global_paths/path_2.csv
patrick-5546: Changes I made to the interface
- Takes the relative file path as first positional argument
- Interval spacing is an optional argument, `-i/--interval_spacing`, default 30.0
- Prints out global path to terminal
- Bring up documentation with `-h`

local_pathfinding/node_mock_global_path.py
```diff
@@ -0,0 +1,220 @@
...
+        # append the destination
+        global_path.waypoints.append(HelperLatLon(latitude=lat2, longitude=lon2))
+
+        if dst_file_path is not None:
+            # write to a new file
+            with open(dst_file_path, "w") as file:
+                writer = csv.writer(file)
+                writer.writerow(["latitude", "longitude"])
+                for waypoint in global_path.waypoints:
+                    writer.writerow([waypoint.latitude, waypoint.longitude])
```
patrick-5546: I see that you now write the global path to the file path. What do you see as the usecase for this? In my mind the node would only need to read from the files: the parameter could be used to quickly/temporarily tests a new path, and if we want to update or create a new file it makes sense to me for this to be a manual choice/process. 

Also with how it's currently implemented most of the time it will be writing the identical contents back into the file (for example when used in the plotting script)

patrick-5546 (approved):
Once you resolve the comments above feel free to merge

local_pathfinding/node_mock_global_path.py
patrick-5546: The logging part wasn't done yet

local_pathfinding/node_mock_global_path.py
SPDonaghy: My thinking was that it might be nice to see the path, that was generated, in an easy to read and editable format. So I came up with the idea of writing the path back into the source file, as I don't see the use case for preserving that single destination point that was in there previously.

patrick-5546: That is a good point, but I also don't want to overwrite files. Maybe you could append the timestamp to the filename?

SPDonaghy: Also, I needed a csv file that had a nice global path in it already, in order to plot it with the plotting script. So at some point we need to write a path to a csv file, in order to test out how the script plots a path from a csv.

patrick-5546: ~~Also we probably don't want to write to the file when plotting, so you can set `dst_file_path` to `None`~~

SPDonaghy: I don't think generate path will write to the csv over and over because it wont be called once the global path is more than one element, which should be the case after generate_path runs once.

But in that case, self.global path will get updated with the generated path once from generate_path() and then again from the csv file on the next callback run. So I will try to eliminate that redundancy.

SPDonaghy: I like the filename + timestamp idea, I will add that

local_pathfinding/node_mock_global_path.py
```diff
@@ -0,0 +1,220 @@
...
+        if not self._all_subs_active():
+            self._log_inactive_subs_warning()
+
+        file_path = self.get_parameter("global_path_filepath")._value
+
+        # check when global path was changed last
+        path_mod_tmstmp = time.ctime(os.path.getmtime(file_path))
+
+        # Only publish if the path has changed
+        if (self.file_path != file_path) or (path_mod_tmstmp != self.path_mod_tmstmp):
```
patrick-5546: One thing I realized is that we may still want publish a new path if the current position from the GPS has changed a lot. This doesn't really matter right now since we're hardcoding it, but would be applicable once we actually read it from the GPS msg. Can implement in this PR or add a TODO comment

local_pathfinding/node_mock_global_path.py
patrick-5546: Can log info that as well

global_paths/plot_global_path.py
```diff
@@ -0,0 +1,99 @@
...
+from typing import Dict, List, Tuple
+
+import plotly.graph_objects as go
+from custom_interfaces.msg import HelperLatLon
+
+from local_pathfinding.node_mock_global_path import MOCK_GPS, MockGlobalPath
+
+
+def main():
+    parser = argparse.ArgumentParser()
```
patrick-5546: Add an optional argument for position, so that we can generate a path from anywhere? Defaults to MOCK_GPS

local_pathfinding/node_mock_global_path.py
SPDonaghy: Oh sorry I thought you meant something else, I understand now

SPDonaghy:
Notes from call:

- [x] CLI arg for whether we want to write it to a new file with the timestamp appended

- [x] interpolate between first point of given path and GPS if they are really far apart

- [x] param to decide if we want to write global path to a filefor when called by the node

- [x] add arg  to plotter for the current position.

SPDonaghy:
@patrick-5546 I  added some new features, it would be great to know your thoughts!

Major changes:

- Path interpolation function: This function analyzes the given path, including the starting position, and fills in any gaps (where waypoints are too far apart) with a subpath, using the same generate_path function
- interval spacing function: a helper function for the interpolator, which returns an array of all the actual interval lengths for a given path
- Purge script for deleting any generated path files in the global_paths directory
- Instead of a service, in order to force the path to update, from the command line, you can just set the parameter `force` to true. The checks in the callback will be bypassed, the path updated, and then the param is set to false again.

SPDonaghy:
More Changes:

Implemented a Path Builder GUI with Flask. Basic functions include:

- Export to CSV
- Clear Path
- 3D Plot (which exports to a csv in the process)

I am a little behind on documentation for this stuff, so I'll be sure to catch up on that. 
Also one persistent issue is I am not able to import plot_global_path.py (from the global_paths folder) in the path_builder.py module. Not sure why, so I have a workaround of pasting in the functions I needed, for now.

FireBoyAJ24:
All your test cases rot variables will need to be in ints to match the type set in the custom interfaces.

SPDonaghy:
> All your test cases rot variables will need to be in ints to match the type set in the custom interfaces.
@FireBoyAJ24 

Were you still going to make a new pr for that? Or would you like me to?

global_paths/path_builder/path_builder.py
patrick-5546, line 1: The interpolation functionality is not implemented yet right? Clicking on the button doesn't do anything, and I don't see any references to interpolation here

local_pathfinding/node_mock_global_path.py
```diff
@@ -0,0 +1,449 @@
...
+        )
+
+        # Attributes
+        self.gps = MOCK_GPS  # TODO Remove when NET publishes GPS
+        self.path_mod_tmstmp = None
+        self.file_path = None
+
+    # Subscriber callbacks
+    def gps_callback(self, msg: GPS):
+        """Store the gps data and check if the global path needs to be updated."""
```
patrick-5546: Docstring needs updating

local_pathfinding/node_mock_global_path.py
```diff
@@ -0,0 +1,449 @@
...
+        to a new csv file in the same directory as the source csv file.
+
+        Global path can be changed by modifying mock_global_path.csv or modifying the
+        global_path_filepath parameter.
+
+        Args:
+            gps_call (bool, optional):
+                Whether the callback was called by the gps callback.
+            force (bool, optional):
+                Whether to force the callback to run, without any checks
```
patrick-5546: Can remove the force argument right, because in the function you're using the parameter

patrick-5546: Or you can remove `gps_call` and in the gps callback call with `force=True`

global_paths/plot_global_path.py
patrick-5546, line 1: Would be good to have a README to explain how to use these scripts and interact with the mock global pathfinding node

local_pathfinding/node_mock_global_path.py
```diff
@@ -0,0 +1,449 @@
...
+        force = self.get_parameter("force")._value
+
+        # Only publish path if the path has changed or gps has changed by more than gps_threshold
+        if (
+            force
+            or gps_call
+            or (self.file_path != file_path)
+            or (path_mod_tmstmp != self.path_mod_tmstmp)
+        ):
+            self.set_parameters([rclpy.Parameter("force", rclpy.Parameter.Type.BOOL, False)])
```
patrick-5546: Update parameters and instance variables (that were used in boolean above) at the bottom

local_pathfinding/node_mock_global_path.py
```diff
@@ -0,0 +1,449 @@
...
+        # check if the global path has been forced to update by parameter change in CLI
+        force = self.get_parameter("force")._value
+
+        # Only publish path if the path has changed or gps has changed by more than gps_threshold
+        if (
+            force
+            or gps_call
+            or (self.file_path != file_path)
+            or (path_mod_tmstmp != self.path_mod_tmstmp)
+        ):
```
patrick-5546: Invert the boolean and exit early so that you don't have such a large if statement

local_pathfinding/node_mock_global_path.py
```diff
@@ -0,0 +1,449 @@
...
+                msg = global_path
+
+            # publish global path
+            self.global_path_pub.publish(msg)
+            self.get_logger().info(
+                f"Publishing to {self.global_path_pub.topic}: {MockGlobalPath.path_to_dict(msg)}"
+            )
+
+    @staticmethod
+    def generate_path(
```
patrick-5546: Functions that are used outside of this node/file can be changed to global functions

local_pathfinding/node_mock_global_path.py
```diff
@@ -0,0 +1,449 @@
...
+                    f"{global_path.waypoints[0].latitude:.4f}, "
+                    f"{global_path.waypoints[0].longitude:.4f}"
+                )
+
+                write = self.get_parameter("write")._value
+                if write:
+                    self.get_logger().info("Writing generated path to new file")
+
+                msg = MockGlobalPath.generate_path(
+                    dest=global_path.waypoints[0],
```
patrick-5546, line 369: If you pass in `waypoints` instead of the first element of `waypoints`, then you wouldn't need to handle 2 `dest` types in `generate_path`, right?

local_pathfinding/node_mock_global_path.py
```diff
@@ -0,0 +1,449 @@
...
+
+        if isinstance(dest, list):
+            lat2 = dest[0].latitude
+            lon2 = dest[0].longitude
+        else:
+            lat2 = dest.latitude
+            lon2 = dest.longitude
+
+        distance = meters_to_km(GEODESIC.inv(lats1=lat1, lons1=lon1, lats2=lat2, lons2=lon2)[2])
+        n = np.ceil(distance / interval_spacing)
```
patrick-5546: So if distance is 25 and interval_spacing is 10, we would get n=3 intermediary points. Then global path would be `pos, i1, i2, i3, dest`, and the distance between waypoints would be 6.25.

Am I understanding this correctly? If so, should n be floor not ceil (not accounting for floating point arithmetic)?

local_pathfinding/node_mock_global_path.py
```diff
@@ -0,0 +1,449 @@
...
+        """Writes the global path to a new, timestamped csv file.
+
+        Args:
+            file_path (str):
+                The filepath to the csv file containing the global path,
+            global_path (Path):
+                The global path to write to file
+        """
+        if file_path == "":
+            raise ValueError("file_path must be specified when write is True")
```
patrick-5546: I don't think this is needed because of the check immediately after

local_pathfinding/node_mock_global_path.py
```diff
@@ -0,0 +1,449 @@
...
+
+        # check if file_path is a valid file path
+        if "global_paths" not in file_path or not os.path.isdir(os.path.dirname(file_path)):
```
patrick-5546: If we want it to be in global_paths should we check if the directory name ends in `global_paths`?

launch/main_launch.py
patrick-5546: Is this documented anywhere?

global_paths/plot_global_path.py
```diff
@@ -0,0 +1,125 @@
...
+
+    with open(file_path, "r") as file:
+        reader = csv.reader(file)
+        # skip header
+        reader.__next__()
+        for row in reader:
+            lats.append(float(row[0]))
+            lons.append(float(row[1]))
+
+    if len(lats) < 2:
```
patrick-5546: If this is supposed to reflect what happens in `global_path_callback`, need to add the case where the waypoints are too far apart. But to reduce complexity maybe we can remove this if statement and only plot the waypoints in the file. Creating the path can be done by path_builder

global_paths/path_builder/path_builder.py
patrick-5546, line 1: Maybe we can combine this file and `plot_global_path.py` and use CLI arguments to control what is done. This would solve your import problem as well

global_paths/plot_global_path.py
patrick-5546, line 1: Also would be nice to have a requirements.txt file for the extra dependencies needed (is it just flask?)

patrick-5546 (commented):
WOW this is really great stuff Sean! Super impressed with path_builder and how you implemented my previous feedback

global_paths/path_builder/path_builder.py
SPDonaghy: Thank-you! I kinda got carried away with it lol. The interpolation works now, with my latest commit!

global_paths/plot_global_path.py
SPDonaghy: It is just flask yes

launch/main_launch.py
SPDonaghy: I'm not sure how to address this/ should I add it to the issue?

local_pathfinding/node_mock_global_path.py
SPDonaghy: something like this? `if (path_mod_tmstmp != self.path_mod_tmstmp or self.file_path != file_path) or force:`?

I removed the gps_call arg, and just set force to true before calling the this callback from the gps callback

local_pathfinding/node_mock_global_path.py
SPDonaghy: you are right, I had an idea to make it accept a path as the destination early on, for use in the interpolator, but didn't end up using that. So I will actually get rid of the option for a list arg as its never used.

local_pathfinding/node_mock_global_path.py
SPDonaghy: > Or you can remove `gps_call` and in the gps callback call with `force=True`

I will do this!

local_pathfinding/node_mock_global_path.py
SPDonaghy: For sure. Also, do you think its neccessary to check isdir as well?

local_pathfinding/node_mock_global_path.py
SPDonaghy: Yes, I believe you're right, my instinct was to use ceil, but we actually need n intervals so n-1 points is correct.

local_pathfinding/node_mock_global_path.py
patrick-5546: I was thinking something like this:
```python
if not force and file_path == self.file_path and path_mod_tmstmp == self.path_mod_tmstmp:
    return
```

local_pathfinding/node_mock_global_path.py
patrick-5546: Yes that is good, can only write to a directory that exists

launch/main_launch.py
patrick-5546: Yea I don't think we have an issue for part 2 yet right

SPDonaghy: No, I will make one

global_paths/plot_global_path.py
SPDonaghy: I like that. I have removed the if, and this function will just get the lats an lons from a file and return them.

PR #66, opened 2023-12-08
Normalizing Each Objective Class's motionCost()

### LINKED ISSUE #33: Custom Weights for Different Optimization Objective
### Purpose
In Raye, we used to make up the cost values for different optimization objectives with @jamenkaye help. This time around we want to ensure our optimization objective weights are appropriate and the summation of optimization objectives. This will allow us to ensure that we have considered the relation between the objectives fully and systematically. 
### Description

1. Research and provide the best method for determining the weights of each optimization objective. You can write up a methodology on confluence for this part. It may provide the simulation team with a bit of insight into how we will use their simulation. 
2. Determine the best method to combine different optimization objectives.

We can ignore the speed objective for now since it is not ready. However, you can still look at the [possible methods](https://ubcsailbot.atlassian.net/wiki/spaces/prjt22/pages/1860403201/How+Do+We+Calculate+the+Optimization+Objectives#Objective-3%3A-Optimum-Speed-of-The-Sailbot-at-its-Initial-Position) to see how we will try to create the speed objective.  
### Resources

- https://ompl.kavrakilab.org/optimizationObjectivesTutorial.html
- https://ompl.kavrakilab.org/classompl_1_1base_1_1MultiOptimizationObjective.html

### LINKED ISSUE #290: Custom Weights for Different Optimization Objective
### Purpose
In Raye, we used to make up the cost values for different optimization objectives with @jamenkaye help. This time around we want to ensure our optimization objective weights are appropriate and the summation of optimization objectives. This will allow us to ensure that we have considered the relation between the objectives fully and systematically. 
### Description

1. Research and provide the best method for determining the weights of each optimization objective. You can write up a methodology on confluence for this part. It may provide the simulation team with a bit of insight into how we will use their simulation. 
2. Determine the best method to combine different optimization objectives.

We can ignore the speed objective for now since it is not ready. However, you can still look at the [possible methods](https://ubcsailbot.atlassian.net/wiki/spaces/prjt22/pages/1860403201/How+Do+We+Calculate+the+Optimization+Objectives#Objective-3%3A-Optimum-Speed-of-The-Sailbot-at-its-Initial-Position) to see how we will try to create the speed objective.  
### Resources

- https://ompl.kavrakilab.org/optimizationObjectivesTutorial.html
- https://ompl.kavrakilab.org/classompl_1_1base_1_1MultiOptimizationObjective.html

akash-venkateshwaran:
### Description

- Resolves UBCSailbot/sailbot_workspace#290 

- Added two functions in the parent class: (i) Creates random sample states; (ii) Compares every pair in that sample and picks the maximum motionCost()
- Inherited these functions in each Objective class to find the max_motionCost for each objective function
- Modified the motionCost() function of each Objective class to normalise its cost using max_motionCost

### Verification

- Verified using a simple setup example problem
- Haven't included unit tests

### Resources

-


local_pathfinding/objectives.py
```diff
@@ -64,14 +104,21 @@ class DistanceObjective(Objective):
...
     """
 
-    def __init__(self, space_information, method: DistanceMethod, reference=cs.LatLon(0, 0)):
-        super().__init__(space_information)
+    def __init__(
+        self,
+        space_information,
+        method: DistanceMethod,
+        reference: cs.LatLon = cs.LatLon(0, 0),
+        num_samples: int = 100,
```
patrick-5546: If you don't think `num_samples` won't change dynamically, we can remove this argument and just hardcode when calling the super's init function

```python
super().__init__(space_information, num_samples=100)
```

patrick-5546 (commented):
I added some tests in `test_distance_objective` that we should use for the other objectives:
- `sample_states()`: commented out checking that all states are valid, which I think is what we want right?
- `find_maximum_motion_cost()`
    - Implicitly test by comparing `max_motion_cost` to a hardcoded value
    - Tried to test `find_maximum_motion_cost()` directly but was not able to get the type of `states` correct so I commented it out for now
        - Update: fixed and uncommented

If you want to merge this PR today and aren't able to finish the tests, I'm ok with merging this PR and working on the tests in another PR. Just leave some `# TODO: <message>` comments

akash-venkateshwaran:
> I added some tests in `test_distance_objective` that we should use for the other objectives:

> 
> * `sample_states()`: commented out checking that all states are valid, which I think is what we want right?
> * `find_maximum_motion_cost()`
>   
>   * Implicitly test by comparing `max_motion_cost` to a hardcoded value
>   * Tried to test `find_maximum_motion_cost()` directly but was not able to get the type of `states` correct so I commented it out for now
>     
>     * Update: fixed and uncommented
> If you want to merge this PR today and aren't able to finish the tests, I'm ok with merging this PR and working on the tests in another PR. Just leave some `# TODO: <message>` comments

Thanks for adding the unit tests.
Yes we want to make sure all the sampled states are valid which means they are not in the collision area
Great! So now I have to repeat these tests for other objective functions right?

local_pathfinding/objectives.py
akash-venkateshwaran: Yup, num_sample won't change dynamically but I think it will change for each objective function. A more complicated objective function would require more samples to find a close estimate of the max_motionCost. Sure, we can hardcode it.

patrick-5546:
> Great! So now I have to repeat these tests for other objective functions right?

Yes and ensure that they are the sampled states are valid. The validity checker currently is just a placeholder but it will be good to set this up now anyways.

test/test_objectives.py
```diff
@@ -20,20 +22,54 @@
...
+            state = ob.State(space)
+            state().setXY(*xy)
+            states.append(state)
+        max_cost = max(
+            distance_objective.ompl_path_objective.motionCost(s1(), s2()).value()
+            for s1, s2 in itertools.combinations(iterable=states, r=2)
+        )
+        assert distance_objective.find_maximum_motion_cost([s() for s in states]) == pytest.approx(
+            max_cost / distance_objective.max_motion_cost
+        )
```
patrick-5546: What do you think about adding an explicit test for the other objectives? Only need to test one method per objective

test/test_objectives.py
```diff
@@ -20,20 +22,54 @@
...
         method,
     )
-    assert distance_objective is not None
+
+    # test sample_states()
+    num_samples = 3
+    sampled_states = distance_objective.sample_states(num_samples)
+    assert len(sampled_states) == num_samples
+    # for state in sampled_states:
+    #     assert ompl_path.is_state_valid(state)
```
patrick-5546: We discussed this a while ago, but I think sample_states should return valid states. Then you can uncomment these 2 lines from all the new tests

patrick-5546 (commented):
It seems like sometimes the normalized cost is greater than 1: https://github.com/UBCSailbot/local_pathfinding/actions/runs/7688802043/job/20950435771

test/test_objectives.py
akash-venkateshwaran: Sure, but it will fail cause the validity check is not implemented in the sampler. I think Sean has to first define the target boats and islands for us to implement this. Let me know if I am wrong. It can easily be added I just have to add that check before appending to the list within sampler function.

test/test_objectives.py
akash-venkateshwaran: Sure but I didn't understand why we are doing this. I think we are checking if find_maximum_motion_cost() gives the same output as max() for three samples. But find_maximum_motion_cost() itself computes it using max() function right?

test/test_objectives.py
patrick-5546: You can use the function in this commented out code, `is_state_valid()`

test/test_objectives.py
patrick-5546: Yea this seems unnecessary, removed

test/test_objectives.py
patrick-5546: Yea no need to do this as we discussed, the cost doesn't take into account the state validity anyways

local_pathfinding/objectives.py
```diff
@@ -40,19 +41,64 @@ class Objective(ob.StateCostIntegralObjective):
...
+            state = self.space_information.getStateSpace().allocState()
+            sampler.sampleUniform(state)
+            sampled_states.append(state)
+
+        return sampled_states
+
+    def capping_motionCost(self, value):
+        if value > 1:
+            value == 1
+        return value
```
patrick-5546: Let's make a function that just returns the normalized cost with capping. So normalize the value with `self.max_motion_cost` before capping and returning

local_pathfinding/objectives.py
```diff
@@ -40,19 +41,64 @@ class Objective(ob.StateCostIntegralObjective):
...
     """
 
-    def __init__(self, space_information):
+    def __init__(self, space_information, num_samples: int):
         super().__init__(si=space_information, enableMotionCostInterpolation=True)
         self.space_information = space_information
 
+        states = self.sample_states(num_samples)
+        self.max_motion_cost = 1.0
+        self.max_motion_cost = self.find_maximum_motion_cost(states)
```
patrick-5546, line 82: I forgot, is there a reason why you initialize it to 1 first? Would it be possible to do this inside `find_maximum_motion_cost()`?

local_pathfinding/objectives.py
patrick-5546: Also don't forget the docstring

local_pathfinding/objectives.py
akash-venkateshwaran: Yes, we divide the motionCost() by max_motion_cost. So initially when we don't know the max_motion_cost we divide it as 1. Then we sample states and find the max_motion_cost. Finally normalise using the new max_motion_cost we obtain from the sample

patrick-5546 (approved):

local_pathfinding/objectives.py
patrick-5546: Akash has done this

local_pathfinding/objectives.py
patrick-5546: Added a comment to explain this

patrick-5546:
As discussed, just need to resolve merge conflicts:
- `git merge origin main`
- Add speed objective
- Normalize speed objective
- Update weights

Issue #28, opened 2023-09-22
Create interpolation function


jamenkaye:
### Purpose

We want to be able to calculate how long the boat will take to travel a given path in known wind conditions. To do this, we need to know the boat speed, which depends on the "[point of sail](https://ubcsailbot.github.io/docs/main/reference/sailing/points_of_sail/)" and the windspeed.

### Description

I've come up with a table of estimates for boat speed, based on the windspeed and sailing angle. This task is to come up with a function to interpolate this data:

Windspeed | 0 deg | 20 deg | 30 deg | 45 deg | 90 deg | 135 deg | 180 deg
-- | -- | -- | -- | -- | -- | -- | --
0      | 0 | 0 | 0 | 0 | 0 | 0 | 0
9.3   | 0 | 0 | 0.4 | 1.1 | 3.2 | 3.7 | 2.8
18.5 | 0 | 0.3 | 1.9 | 3.7 | 9.3 | 13.0 | 9.2
27.8 | 0 | 0.9 | 3.7 | 7.4 | 14.8 | 18.5 | 13.0
37.0 | 0 | 1.3 | 5.6 | 9.3 | 18.5 | 24.1 | 18.5

The exact values in the table _**or the dimensions of the table might get changed later**_ so make sure your implementation is flexible enough to allow this easily.

The function arguments will be:
- [True Windspeed](https://ubcsailbot.github.io/docs/main/reference/sailing/miscellaneous/#:~:text=angle%20conventions%20are.-,True%2C%20Apparent%2C%20and%20Boat%20Wind,-%C2%B6) in km/h (Make sure function header states that it's True Wind)
- Sailing angle in degrees (Eg, 0 degrees means the boat is in Irons, 45 degrees means the boat is on Close Hauled, 180 degrees means the boat is on a Run)

The function return will be:
- Boat speed in km/h

There are tons of different mathematical approaches to interpolation. I would suggest starting with something like the simple approach shown below, but you are free to look into more complicated interpolations if you like :)

1. Suppose function inputs are 60 degrees, 12 knots.
2. Lets look at the 10-knot row of the table first. $2 \cdot \frac{90-60}{90-45} + 5 \cdot \frac{60-45}{90-45} = 3$ so our first intermediate value is 3 knots.
3. Now lets look at the 15 knot row of the table. $4 \cdot \frac{90-60}{90-45} + 8 \cdot \frac{60-45}{90-45} = 5.33$, which is our second intermediate value.
[...]

PR #67, opened 2023-12-20
Creating a Skeleton Speed Objective

### LINKED ISSUE #26: Create a Speed Objective
# Purpose
The speed objective is used to determine the sailbot's cost when travelling at a particular "[point of sail](https://ubcsailbot.github.io/docs/main/reference/sailing/points_of_sail/)" and windspeed. This will penalize the sailbot from going above or below a particular speed. It safeguards the sailbot from possibly damaging the sailbot at high speeds or becoming stationary at lower speeds.

# Description

1. Determine how we will store the polar plot data and interpolate the sailbot speeds when our windspeed and angles are in between the discrete inputs. 
a. There are a number of ways to interpolate data points - Linear, Polynomial and Spline interpolation. Therefore, we will need to evaluate each method. 
2. Once we determine the initial speed at that point, determine the cost of the sailbot if the sailbot is within a particular speed range or not.
b. Do we use an inverse bell shape curve or linearly determine the cost? We will need to evaluate different methods of determining the costs.

Ensure you note the purpose of the functions, and what are the inputs and outputs of each of the functions. Note any restrictions on the inputs of the functions. The class function should be added to the path_objective.py file.

# Summary

- [ ] Construct a skeleton code of the speed objective
- [x] #28 
- [ ] Evaluate a suitable method for determining the cost of the sailbot at that particular speed.

# Resources
- https://ubcsailbot.atlassian.net/wiki/spaces/prjt22/pages/edit-v2/1860403201

FireBoyAJ24:
### Description

The pull request will add a speed objective and its skeleton test cases to complete adding all the optimization objectives for  Polaris. The pull requests come from the first task item in issue #26 

### Verification

- [x] Check whether all the speed objective functions are not producing an error.
- [x] Ensure all the skeleton speed objective test cases are passing

### Resources

- #26
- #28
- #42


local_pathfinding/objectives.py
```diff
@@ -380,8 +399,111 @@ def is_angle_between(first_angle: float, middle_angle: float, second_angle: floa
...
+    @staticmethod
+    def get_continuous_cost(speed: float) -> float:
+        """Generates the cost associated with the speed of the boat.
+
+        Args:
+            speed (float): The speed of the boat in m/s
+        """
+        try:
+            cost = 1 / math.sin(math.pi * speed / 25) - 0.5
+            return 10000 if cost > 10000 else cost
```
jamenkaye: I think you could say, `return min(10000, cost)` right?

jamenkaye (approved):
looks good to me!

PR #70, opened 2024-01-16
Global Path HTTP Module

### LINKED ISSUE #69: Mock Global Path Part 2: WEB & NET
### Purpose

This is the second part to the global path module implementation (the right side of the diagram below)
### Description

This Global Path Module will receive the global path from WEB via POST request and then send it via POST to the Remote Transceiver.
Some features may overlap with the part one module, and the two modules may be combined at a later time.

### Resources
General structure of parts 1 and 2 of the mock global path implementation.

image

SPDonaghy:
### Description

- Resolves #69 

Aside from the Issue, I made some small changes, here and there, to my work on the previous PR, just some renaming and set write to False in the mgp test file.

### Verification

- Pytests for new functions
- Created mock HTTP server for endpoint for NET to test out the POST request made by the module, to send the global path to NET.
- HTTP request test coverage also with Pytests

### Resources

-


local_pathfinding/global_path.py
```diff
@@ -0,0 +1,247 @@
...
+        print(f"HTTP Error: {http_error.code}")
+    except URLError as url_error:
+        print(f"URL Error: {url_error.reason}")
+    except Exception as e:
+        print(f"An unexpected error occurred: {e}")
+
+    lat_lon = position["lat_lon"]
+    pos = HelperLatLon(latitude=lat_lon["latitude"], longitude=lat_lon["longitude"])
+
+    return pos
```
patrick-5546: @SPDonaghy if you run the website (uncomment docker-compose.website.yml in devcontainer.json then rebuild dev container) you will get a response from the endpoint.

However, currently it is empty: 
```
{'success': True, 'data': []}
```

@UBCSailbot/website can you populate the GPS endpoint with something, or give us what it should look like when populated?

@SPDonaghy maybe if we don't get a response or it is empty we can population `position` with a constant in our file?

SPDonaghy: @patrick-5546  Sure! I was also thinking if we don't get a response or if its empty we would just print to console and keep trying the GET request? Since the path we generate with a constant wont be too meaningful anyways.

jahn18: Here's an example of what the GPS endpoint returns when populated:

```
{
	success: true,
	data: [
		{
			latitude: 49.376629216076275,
			longitude: -123.27266500802254,
			speed: 0,
			heading: 0,
			timestamp: "2024-01-19T20:42:50.509Z"
		},
		{
			latitude: 49.37711663428484,
			longitude: -123.27156381625609,
			speed: 0,
			heading: 0,
			timestamp: "2024-01-19T20:42:50.510Z"
		}
	]
}
``` 

FYI: The array is sorted based on the timestamp, so if you want the current position of the boat, you'd need to get the last element.

----- 

There's also the option to populate the endpoint locally by running our tests:

1. Access the website container terminal on Docker
2. Go into the directory "tests", e.g. `cd tests` 
3. Run `npm install`
4. Run `npm run test-tag @gps` 

This should write some sample gps data into the DB

SPDonaghy: I ran the website by uncommenting the docker-compose.website.yml file and rebuilding the container, but I am having trouble populating the endpoint with data because npm is not a recognized command. Is there something else I need to change to make sure npm is installed?

patrick-5546: You have to run it in the website container from docker desktop

image

local_pathfinding/node_mock_global_path.py
```diff
@@ -42,7 +42,7 @@ def generate_path(
     with the name of the original file, appended with a timestamp.
 
     Args:
-        dest (Union[HelperLatLon, list[HelperLatLon]]): The destination point or partial path
+        dest (list[HelperLatLon]): The destination point or partial path
```
patrick-5546: It's just a point right? Not a partial path / list anymore

patrick-5546 (commented):
Great work! What do you think about importing common functions from `global_path.py` in `node_mock_global_path.py` instead of vice versa (how it currently is) so that it can be standalone. I think this would also require converting the ROS parameters you added for mgp_main to constants.

Also how about adding some functionality in `node_mock_global_path.py` to publish or post depending on a ros parameter

Let's discuss about integration with WEB and NET in the soft-wide meeting on Saturday (I added a discussion topic in the meeting notes, feel free to add to it)

SPDonaghy:
> Great work! What do you think about importing common functions from `global_path.py` in `node_mock_global_path.py` instead of vice versa (how it currently is) so that it can be standalone. I think this would also require converting the ROS parameters you added for mgp_main to constants.
> 
> Also how about adding some functionality in `node_mock_global_path.py` to publish or post depending on a ros parameter
> 
> Let's discuss about integration with WEB and NET in the soft-wide meeting on Saturday (I added a discussion topic in the meeting notes, feel free to add to it)

Yeah it definitely makes more sense to do it the other way around, I agree. For sure, lets discuss it more in the meeting on Sat.

local_pathfinding/node_mock_global_path.py
SPDonaghy: yes, with the intro of the _interpolate_path() function, generate_path only works with a destination point now.

patrick-5546:
Following up on our discussion today, plan for tests is to read from DB directly instead of `get_server.py`, and to use `post_server.py` until NET has implemented their endpoint (let's create an issue to remove `post_server.py` once it is) - I mentioned this PR in a sailbot workspace PR that is required for CI to pass after making these changes

patrick-5546:
I'm guessing tests are still failing because ports aren't being forwarded. In sailbot workspace this is done in devcontainer.json, but that isn't used for running CI (just the compose files are)

SPDonaghy:
> I'm guessing tests are still failing because ports aren't being forwarded. In sailbot workspace this is done in devcontainer.json, but that isn't used for running CI (just the compose files are)

Sorry I'm a little confused, what action is needed from me?

patrick-5546:
It was a note for myself, will fix it when I have time (sometime this week)

patrick-5546 (approved):
Finally got CI to pass. It would have been much simpler with a monorepo. Have you created an issue to use the NET endpoint once it's implemented yet?

PR #74, opened 2024-01-23
Added the Website Interface Node (Part 1)

### LINKED ISSUE #62: Website Interface Node
### Purpose

We want to publish local pathfinding data to the website for visualization and debugging

### Description

- In node_navigate.py we want to publish to LPath Data (add to the node_navigate python file). 
- Want to create a publisher, its timer, and callback function 
- In the callback function, want to initialize a message of type LPath Data, initialize its variables, and publish it
- Covert LocalPath.waypoints (Found in LocalPath) to be published as a LPath Data message
- LocalPath is an instance variable of node_navigate, so to get LocalPath we need to do self.local_path

### Resources

- LPath data is documented here: https://github.com/UBCSailbot/custom_interfaces
- LPath Data message: https://github.com/UBCSailbot/custom_interfaces/blob/main/msg/LPathData.msg
- Local Path Defintion: https://github.com/UBCSailbot/local_pathfinding/blob/main/local_pathfinding/local_path.py
- Node_navigate Node: https://github.com/UBCSailbot/local_pathfinding/blob/main/local_pathfinding/node_navigate.py
    - Use desired heading publisher as reference. Infrastructure-wise, it should be identical so just do this for LPath Data

FireBoyAJ24:
### Description
- Resolves UBCSailbot/local_pathfinding#62

- Created the publisher, timer and callback function
- Adjusted the get_waypoints function so that the XY coordinates is converted to latlon coordinates

### Verification

- [ ] Verify by using debug statements
- [ ] Verify by checking the log to ensure LPathData is added


local_pathfinding/node_navigate.py
```diff
@@ -71,11 +78,17 @@ def __init__(self):
         self.desired_heading_pub = self.create_publisher(
             msg_type=DesiredHeading, topic="desired_heading", qos_profile=10
         )
+        self.LPath_pub = self.create_publisher(
+            msg_type=LPathData, topic="local_path", qos_profile=10
```
jamenkaye: I think the topic name should be `local_path_data`, since we plan to add more than just the path, and this matches the [diagram](https://github.com/UBCSailbot/custom_interfaces?tab=readme-ov-file#project-wide-external-interfaces) in custom interfaces

local_pathfinding/node_navigate.py
```diff
@@ -141,6 +154,35 @@ def get_desired_heading(self) -> float:
         # TODO: create function to compute the heading from current position to next local waypoint
         return 0.0
 
+    def LPath_callback(self):
+        """Get and publish the local path."""
+
+        waypoints = self.get_LPath()
+
+        local_path = Path(waypoints=waypoints)
```
jamenkaye: It might be nice to rename the `waypoints` variable to `current_waypoints` or something, so line 160 is less confusing. I might just be traumatized from trying to learn C++ tho so up to you haha

local_pathfinding/node_navigate.py
```diff
@@ -141,6 +154,35 @@ def get_desired_heading(self) -> float:
...
+
+        msg = LPathData(local_path=local_path)
```
jamenkaye: Same thing occurs here, if you feel inclined to change the name of `local_path` to `current_local_path` or something

local_pathfinding/node_navigate.py
```diff
@@ -141,6 +154,35 @@ def get_desired_heading(self) -> float:
...
+    def get_LPath(self):
+        """Get the local path.
+
+        Returns:
+            List[Tuple[float, float]]: The local path if all subscribers are active, else an empty
+                list.
+        """
+        if not self._all_subs_active():
+            self._log_inactive_subs_warning()
+            return -1.0
```
jamenkaye: Good idea to check for this error! What do we want the code to do in the case of inactive subs though? I tried running it, and it seems like if we return `-1.0` in place of a path, this causes a fatal erroranyway.

If we want this to stop code execution, maybe we should log it with a severe? Or if we want the code to keep running, maybe we should log a warning and return (0.0, 0.0) or something that should still be valid, but will look fishy?

jamenkaye (commented):
Looks good overall! I left a few comments. I also confirmed that the new `local_path` topic shows up with in a terminal window by running `ros2 topic echo local_path`

local_pathfinding/node_navigate.py
FireBoyAJ24: I will then return a HelperLatLon(0.0,0.0) when no subscribers are active.

local_pathfinding/node_navigate.py
```diff
@@ -141,6 +171,23 @@ def get_desired_heading(self) -> float:
...
+            List[Tuple[float, float]]: The local path if all subscribers are active, else an empty
+                list.
+        """
+        if not self._all_subs_active():
+            self._log_inactive_subs_warning()
+            return [HelperLatLon(0.0, 0.0)]
+
+        self.local_path.update_if_needed(
+            self.gps, self.ais_ships, self.global_path, self.filtered_wind_sensor
+        )
```
patrick-5546: Don't update local path, this is already done in the desired heading callback

local_pathfinding/node_navigate.py
```diff
@@ -141,6 +171,23 @@ def get_desired_heading(self) -> float:
...
 
+    def get_lpath(self):
+        """Get the local path.
+
+        Returns:
+            List[Tuple[float, float]]: The local path if all subscribers are active, else an empty
+                list.
+        """
+        if not self._all_subs_active():
+            self._log_inactive_subs_warning()
```
patrick-5546: Don't log inactive subs, this is already done in the desired heading callback

local_pathfinding/node_navigate.py
```diff
@@ -141,6 +171,23 @@ def get_desired_heading(self) -> float:
         # TODO: create function to compute the heading from current position to next local waypoint
         return 0.0
 
+    def get_lpath(self):
```
patrick-5546: Don't need a separate helper function, just put this implementation in lpath_data_callback directly

local_pathfinding/ompl_path.py
```diff
@@ -87,7 +89,18 @@ def get_waypoints(self) -> List[Tuple[float, float]]:
             return []
 
         solution_path = self._simple_setup.getSolutionPath()
-        waypoints = [(state.getX(), state.getY()) for state in solution_path.getStates()]
+        referenceLatlon = cs.LatLon(*self.state.goal_state)
```
patrick-5546: You need to complete the TODO in OMPLPathState: `reference_latlon` should be `local_path_state.global_path[-1]`

patrick-5546 (changes requested):
In addition to the below, need to fix `test_OMPLPath_get_waypoint`

local_pathfinding/node_navigate.py
```diff
@@ -67,15 +77,23 @@ def __init__(self):
...
+
+        # publisher timers
         pub_period_sec = self.get_parameter("pub_period_sec").get_parameter_value().double_value
         self.get_logger().debug(f"Got parameter: {pub_period_sec=}")
         self.desired_heading_timer = self.create_timer(
             timer_period_sec=pub_period_sec, callback=self.desired_heading_callback
         )
+        self.lpath_data_timer = self.create_timer(
+            timer_period_sec=pub_period_sec, callback=self.lpath_data_callback
+        )
```
patrick-5546, line 96: I think we should publish lpath_data at a lower frequency

FireBoyAJ24: At what frequency should we publish lpath_data?

patrick-5546: Actually sorry can you revert this to `pub_period_sec`, let's let NET determine how often to write to the DB rather than us determining it for us

global_paths/path_2_2024-01-24_22-44-36.csv
patrick-5546, line 1: Did you mean to commit these global paths? I don't see them being used anywhere

local_pathfinding/node_navigate.py
```diff
@@ -26,15 +26,17 @@ class Sailbot(Node):
         global_path_sub (Subscription): Subscribe to a `Path` msg.
         filtered_wind_sensor_sub (Subscription): Subscribe to a `WindSensor` msg.
 
-    Publishers and their timers:
+    Publishers:
         desired_heading_pub (Publisher): Publish the desired heading in a `DesiredHeading` msg.
+
+    Publisher timers:
         desired_heading_timer (Timer): Call the desired heading callback function.
```
patrick-5546: Add new class variables to docstring

local_pathfinding/ompl_path.py
```diff
@@ -34,6 +36,12 @@ def __init__(self, local_path_state: LocalPathState):
         self.start_state = (0.5, 0.4)
         self.goal_state = (0.5, -0.4)
 
+        self.reference_latlon = (
+            local_path_state.global_path[-1]
+            if local_path_state and len(local_path_state.global_path) > 0
+            else HelperLatLon(latitude=0.0, longitude=0.0)
+        )
```
patrick-5546, line 43: Now that I think about it more this is still not right. I'm thinking in LocalPathState we need to find the waypoint on the global path we're heading to (not always the last element), which will be our reference latlon here. Maybe this can be your next task Jay (doesn't have to be tackled in this PR, but let's at least write an issue for properly populating OMPLPathState).

patrick-5546 (approved):

local_pathfinding/ompl_path.py
FireBoyAJ24: Issue #79 has been made from this issue.

PR #77, opened 2024-01-25
Use HelperLatLon instead of LatLon

patrick-5546:
### Description

- Removed `coord_systems.LatLon` as it is redundant with `custom_interfaces.msg.HelperLatLon`
- This is a **breaking change**, may have to resolve merge conflicts with open PR's after merging (so may want to get your PR's in first)
    - However I want to merge this into #74 before continuing
- One difference is that you can't instantiate with `HelperLatLon(*latlon)`
    - Must explicitly name arguments: `HelperLatLon(latitude=latlon[0], longitude=latlon[1])`
    - This is a good practice anyways, just more verbose

@UBCSailbot/pathfinding FYI

### Verification

- [x] Tests task passes

### DIFF
--- a/local_pathfinding/coord_systems.py
+++ b/local_pathfinding/coord_systems.py
@@ -3,21 +3,10 @@
 import math
 from typing import NamedTuple
 
+from custom_interfaces.msg import HelperLatLon
 from pyproj import Geod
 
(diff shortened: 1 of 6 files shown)


jamenkaye (approved):
Looks good to me!

PR #80, opened 2024-01-27
Added logic to handle ros parameter to choose planners

### LINKED ISSUE #71: Add ROS parameter to select planners
### Purpose

The purpose of this feature is to allow users to seamlessly switch between different path planners to test which planners are the most efficient for our use case.

### Description

- Edit [global.yaml](https://github.com/UBCSailbot/sailbot_workspace/blob/main/src/global_launch/config/globals.yaml) file in the sailbot_workspace repository to add another section for switching planners.
- Add documentation for the parameter in the [readme file](https://github.com/UBCSailbot/sailbot_workspace/blob/main/src/global_launch/config/README.md).
- Parse the parameter in the [node_navigate.py file](https://github.com/UBCSailbot/local_pathfinding/blob/main/local_pathfinding/node_navigate.py) in local_pathfinding.

### Resources

- ROS Parameter documentation is here: https://docs.ros.org/en/humble/Concepts/Basic/About-Parameters.html
- globals.yaml file is here: https://github.com/UBCSailbot/sailbot_workspace/blob/main/src/global_launch/config/globals.yaml
- readme.md file is here: https://github.com/UBCSailbot/sailbot_workspace/blob/main/src/global_launch/config/README.md
- node_navigate.py file is here: https://github.com/UBCSailbot/local_pathfinding/blob/main/local_pathfinding/node_navigate.py

Krithik1:
### Description

- Resolves #71 
- Added logic to pipe path_planner ros parameter to `ompl_path.py` file from `node_navigate.py` file via the `local_path.py` file.
- Selects planner based on parameter in `ompl_path.py` file

### Verification

- Log statements are printed for the planner, showing the planner was successfully loaded.

### DIFF
--- a/local_pathfinding/local_path.py
+++ b/local_pathfinding/local_path.py
@@ -30,6 +30,7 @@ class LocalPathState:
         ais_ships: AISShips,
         global_path: Path,
         filtered_wind_sensor: WindSensor,
+        planner: str,
     ):
         """Initializes the state from ROS msgs."""
         if gps:  # TODO: remove when mock can be run
@@ -49,6 +50,8 @@ class LocalPathState:
             self.wind_speed = filtered_wind_sensor.speed.speed
             self.wind_direction = filtered_wind_sensor.direction
 
+        self.planner = planner
+
 
 class LocalPath:
     """Sets and updates the OMPL path and the local waypoints
@@ -60,17 +63,15 @@ class LocalPath:
             to the next global waypoint.
     """
 
-    def __init__(self, parent_logger: RcutilsLogger, planner: str = "rrtstar"):
+    def __init__(self, parent_logger: RcutilsLogger):
         """Initializes the LocalPath class.
 
         Args:
             `parent_logger` (RcutilsLogger): ROS logger of the parent class.
-            `planner` (str): Name of the planner to use. Defaults to "rrtstar".
         """
         self._logger = parent_logger.get_child(name="local_path")
         self._ompl_path: Optional[OMPLPath] = None
         self.waypoints: Optional[List[Tuple[float, float]]] = None
-        self.planner = planner
 
     def update_if_needed(
         self,
@@ -78,6 +79,7 @@ class LocalPath:
         ais_ships: AISShips,
         global_path: Path,
         filtered_wind_sensor: WindSensor,
+        planner: str,
     ):
         """Updates the OMPL path and waypoints. The path is updated if a new path is found.
 
@@ -87,12 +89,11 @@ class LocalPath:
             `global_path` (Path): Path to the destination.
             `filtered_wind_sensor` (WindSensor): Wind data.
         """
-        state = LocalPathState(gps, ais_ships, global_path, filtered_wind_sensor)
+        state = LocalPathState(gps, ais_ships, global_path, filtered_wind_sensor, planner)
         ompl_path = OMPLPath(
             parent_logger=self._logger,
             max_runtime=1.0,
             local_path_state=state,
-            planner=self.planner,
         )
         if ompl_path.solved:
             self._logger.info("Updating local path")
--- a/local_pathfinding/local_path.py
+++ b/local_pathfinding/local_path.py
@@ -60,10 +60,11 @@ class LocalPath:
             to the next global waypoint.
     """
 
-    def __init__(self, parent_logger: RcutilsLogger):
+    def __init__(self, parent_logger: RcutilsLogger, planner: str):
         self._logger = parent_logger.get_child(name="local_path")
         self._ompl_path: Optional[OMPLPath] = None
         self.waypoints: Optional[List[Tuple[float, float]]] = None
+        self.planner = planner
 
     def update_if_needed(
         self,
@@ -81,7 +82,12 @@ class LocalPath:
             `filtered_wind_sensor` (WindSensor): Wind data.
         """
         state = LocalPathState(gps, ais_ships, global_path, filtered_wind_sensor)
-        ompl_path = OMPLPath(parent_logger=self._logger, max_runtime=1.0, local_path_state=state)
+        ompl_path = OMPLPath(
+            parent_logger=self._logger,
+            max_runtime=1.0,
+            local_path_state=state,
+            planner=self.planner,
+        )
         if ompl_path.solved:
             self._logger.info("Updating local path")
             self._update(ompl_path)
--- a/local_pathfinding/local_path.py
+++ b/local_pathfinding/local_path.py
@@ -60,7 +60,13 @@ class LocalPath:
             to the next global waypoint.
     """
 
-    def __init__(self, parent_logger: RcutilsLogger, planner: str):
+    def __init__(self, parent_logger: RcutilsLogger, planner: str = "rrtstar"):
+        """Initializes the LocalPath class.
+
+        Args:
+            `parent_logger` (RcutilsLogger): ROS logger of the parent class.
+            `planner` (str): Name of the planner to use. Defaults to "rrtstar".
+        """
         self._logger = parent_logger.get_child(name="local_path")
         self._ompl_path: Optional[OMPLPath] = None
         self.waypoints: Optional[List[Tuple[float, float]]] = None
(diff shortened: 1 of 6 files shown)


local_pathfinding/node_navigate.py
```diff
@@ -77,14 +78,17 @@ def __init__(self):
             timer_period_sec=pub_period_sec, callback=self.desired_heading_callback
         )
 
+        planner = self.get_parameter("path_planner").get_parameter_value().string_value
+        self.get_logger().debug(f"Got parameter: {planner=}")
```
patrick-5546: Just realized we can't just read parameter in init because updating during runtime won't work. Can you move getting parameters to its own function. This function would get all parameters, and if they changed then store in instance variable and debug log its new value. Then you can call the function both in the init function and desired heading callback.

local_pathfinding/node_navigate.py
```diff
@@ -77,14 +78,17 @@ def __init__(self):
...
+
         # attributes from subscribers
         self.ais_ships = None
         self.gps = None
         self.global_path = None
         self.filtered_wind_sensor = None
 
         # attributes
-        self.local_path = LocalPath(parent_logger=self.get_logger())
+        self.local_path = LocalPath(parent_logger=self.get_logger(), planner=planner)
```
patrick-5546: Rather than making planner an init argument make it an argument of local_path.update_if_needed. And to match with the rest of the code make planner an instance variable of `LocalPathState` and `OMPLPathState`

local_pathfinding/ompl_path.py
```diff
@@ -49,16 +49,19 @@ def __init__(
         parent_logger: RcutilsLogger,
         max_runtime: float,
         local_path_state: LocalPathState,
+        planner: str = "rrtstar",
```
patrick-5546: Don't put default here

local_pathfinding/ompl_path.py
```diff
@@ -97,6 +100,47 @@ def update_objectives(self):
         """
         raise NotImplementedError
 
+    def choose_planner(self, planner: str, si: ob.SpaceInformation) -> ob.Planner:
```
patrick-5546: Make this a global function and put at the end of the file, since it does use any instance variables

local_pathfinding/ompl_path.py
```diff
@@ -97,6 +100,47 @@ def update_objectives(self):
...
+
+        Args:
+            planner (str): Name of the planner to use.
+            si (ob.SpaceInformation): Encapsulates the planning problem to be solved.
+
+        Returns:
+            ob.Planner: Planner to use for the OMPL query.
+        """
+
+        match planner.lower():
```
patrick-5546: Alphabetize

local_pathfinding/ompl_path.py
```diff
@@ -97,6 +100,47 @@ def update_objectives(self):
...
+            case "lazyprmstar":
+                return og.LazyPRMstar(si)
+            case "rrtconnect":
+                return og.RRTConnect(si)
+            case "lbtrrt":
+                return og.LBTRRT(si)
+            case "lazylbtrrt":
+                return og.LazyLBTRRT(si)
+            case _:
+                ou.OMPL_ERROR("Planner-type is not implemented in allocation function.")
```
patrick-5546: Use ros log error, printing what the given planner was and return one of the planners by default

local_pathfinding/ompl_path.py
```diff
@@ -156,9 +200,8 @@ def _init_simple_setup(self) -> og.SimpleSetup:
         simple_setup.setOptimizationObjective(objective)
 
         # set the planner of the simple setup object
-        # TODO: implement and add planner here
-        planner = og.RRTstar(space_information)
-        simple_setup.setPlanner(planner)
+        # TODO: implement and add planner
+        simple_setup.setPlanner(self.choose_planner(self.planner, space_information))
```
patrick-5546: Keep the same structure as before (2 separate lines, it's clearer), just remove the todo

local_pathfinding/ompl_path.py
```diff
@@ -97,6 +100,47 @@ def update_objectives(self):
...
+        """Choose the planner to use for the OMPL query.
+
+        Args:
+            planner (str): Name of the planner to use.
+            si (ob.SpaceInformation): Encapsulates the planning problem to be solved.
+
+        Returns:
+            ob.Planner: Planner to use for the OMPL query.
+        """
+
```
patrick-5546: Remove extra space

local_pathfinding/local_path.py
```diff
@@ -60,10 +60,17 @@ class LocalPath:
...
     """
 
-    def __init__(self, parent_logger: RcutilsLogger):
+    def __init__(self, parent_logger: RcutilsLogger, planner: str = "rrtstar"):
+        """Initializes the LocalPath class.
+
+        Args:
+            `parent_logger` (RcutilsLogger): ROS logger of the parent class.
+            `planner` (str): Name of the planner to use. Defaults to "rrtstar".
+        """
```
patrick-5546: No need for init docstring unless you do something really funky

patrick-5546 (changes requested):

local_pathfinding/ompl_path.py
Krithik1: Made this global function so ros logger no longer accessible

local_pathfinding/node_navigate.py
```diff
@@ -134,17 +131,35 @@ def get_desired_heading(self) -> float:
...
 
         self.local_path.update_if_needed(
-            self.gps, self.ais_ships, self.global_path, self.filtered_wind_sensor
+            self.gps, self.ais_ships, self.global_path, self.filtered_wind_sensor, planner
         )
 
         # TODO: create function to compute the heading from current position to next local waypoint
         return 0.0
 
+    def get_params(self):
```
patrick-5546: Let's do something a bit fancier here. Only when the parameter has changed:
- Save to instance variables (so don't have to return anything)
- Debug log
- [Only when `pub_period_sec` changes] Update timers
    - So you probably want to move the timer initializations to inside this function

After you have made these changes you should verify them by trying to change them while local pathfinding is running:
- In one terminal: `ros2 launch local_pathfinding main_launch.py`
- In another terminal: `ros2 set param navigate_main <param> <value>`

local_pathfinding/node_navigate.py
```diff
@@ -134,17 +131,35 @@ def get_desired_heading(self) -> float:
             float: The desired heading if all subscribers are active, else a number that violates
                 the heading convention.
         """
+        _, planner = self.get_params()
```
patrick-5546: Should go right before `update_if_needed()` (want to check if all subs are active before doing anything)

local_pathfinding/ompl_path.py
```diff
@@ -34,6 +34,9 @@ def __init__(self, local_path_state: LocalPathState):
         self.start_state = (0.5, 0.4)
         self.goal_state = (0.5, -0.4)
 
+        if local_path_state:
+            self.planner = local_path_state.planner
```
patrick-5546: Is this necessary? When is `local_path_state` None?

local_pathfinding/ompl_path.py
```diff
@@ -218,3 +177,44 @@ def is_state_valid(state: ob.SE2StateSpace) -> bool:
...
+        case "rrtsharp":
+            return og.RRTsharp(si)
+        case "rrtstar":
+            return og.RRTstar(si)
+        case "rrtxstatic":
+            return og.RRTXstatic(si)
+        case "sorrtstar":
+            return og.SORRTstar(si)
+        case _:
+            return og.RRTStar(si)
```
patrick-5546: I'm still not completely satisfied with this. Ideally I would want data validation of the planner ros param to occur in `get_params()` node_navigate.py (error log if not valid) but for it to be converted into `og.Planner` here. If you have any ideas feel free to suggest/implement them, I'll keep thinking about this as well

patrick-5546: I attempted to tackle this in my latest commit, what do you think?

patrick-5546 (approved):
lgtm, great work on this PR!

I cleaned up a couple things, take a look at node_navigate.py to see if you like the flow that I was going for

PR #82, opened 2024-01-31
Added linear_interpolation

### LINKED ISSUE #26: Create a Speed Objective

Justin-Prasad:
### Description

- Resolves #26 

Added the interpolation function 

### Verification

- [Added test cases] 

### Resources

N/A

### DIFF
--- a/local_pathfinding/objectives.py
+++ b/local_pathfinding/objectives.py
@@ -466,7 +466,9 @@ class SpeedObjective(Objective):
 
     @staticmethod
     def get_sailbot_speed(heading: float, wind_direction: float, wind_speed: float) -> float:
+        # Get the sailing angle: [0, 180]
         sailing_angle = abs(heading - wind_direction)
+        sailing_angle = min(sailing_angle, 360 - sailing_angle)
 
         # Find the nearest windspeed values above and below the true windspeed
         lower_windspeed_index = max([i for i, ws in enumerate(WINDSPEEDS) if ws <= wind_speed])
--- a/local_pathfinding/objectives.py
+++ b/local_pathfinding/objectives.py
@@ -3,6 +3,7 @@
 import math
 from enum import Enum, auto
 
+import numpy as np
 from ompl import base as ob
 
 import local_pathfinding.coord_systems as cs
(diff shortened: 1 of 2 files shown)


jamenkaye:
@ / Justin-Prasad Running local pathfinding worked when I tried it just now, so I'm guessing some other dependency on your computer was out of date. If you're curious, you could try pulling the latest changes to the main branches of custom interfaces and sailbot workspace, and running the rebuild all task. I also converted our tests to the 4-argument form required, and made sure all the tests passed 👍

jamenkaye:
@patrick-5546 We had to convert the `BOATSPEEDS` table to an numpy array to fix an error from mypy: "Value of type "object" is not indexable". Is this alright, or is there a better solution?

jamenkaye:
@FireBoyAJ24 Just wanted to triple check the conventions you'll use with this function's inputs:
- windspeed: the **true windspeed** measured in km/h
- heading: which way the front of the boat is pointing. 0 degrees means pointing to the north, and angles increase clockwise. Bounded in [-180, 180]
- wind_direction: the direction that the **true wind** is _coming from_, where 0 degrees means wind coming from the north, and angles increase clockwise. Bounded in [-180, 180]. This requires a bit more calculation than the apparent wind measurement that we get from the wind sensors.

patrick-5546:
> Is this alright, or is there a better solution?

Mypy doesn't like how `BOATSPEEDS` has lists that containing different types (int and float) even though Python allows this. Using a numpy array will cast everything to float, or you can just change all `0`'s to `0.0` so that everything is a float.

PR #86, opened 2024-02-11
Logged Inactive Subscribers

### LINKED ISSUE #85: Log inactive subscribers
### Purpose

We'd like to log which subscribers are inactive.

### Description

- Just check all the ROS topics and see if the value is `None`

### Resources

- https://github.com/UBCSailbot/local_pathfinding/blob/66ea6216b625cf897c59f61459663a5009723bd9/local_pathfinding/node_navigate.py#L176

Krithik1:
### Description

- Resolves #85 
- Checks if the 4 subscribers are None. If yes, then a warning is logged.
- Function definition in `node_navigate.py`

### DIFF
--- a/local_pathfinding/node_navigate.py
+++ b/local_pathfinding/node_navigate.py
@@ -174,8 +174,17 @@ class Sailbot(Node):
         return self.ais_ships and self.gps and self.global_path and self.filtered_wind_sensor
 
     def _log_inactive_subs_warning(self):
-        # TODO: log which subscribers are inactive
-        self.get_logger().warning("There are inactive subscribers")
+        """
+        Logs a warning message for each inactive subscriber.
+        """
+        if self.ais_ships_sub is None:
+            self.get_logger().warning("AIS Ships subscriber not active")
+        if self.gps_sub is None:
+            self.get_logger().warning("GPS subscriber not active")
+        if self.global_path_sub is None:
+            self.get_logger().warning("Global Path subscriber not active")
+        if self.filtered_wind_sensor_sub is None:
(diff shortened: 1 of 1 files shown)


local_pathfinding/node_navigate.py
```diff
@@ -174,8 +174,17 @@ def _all_subs_active(self) -> bool:
...
+        Logs a warning message for each inactive subscriber.
+        """
+        if self.ais_ships_sub is None:
+            self.get_logger().warning("AIS Ships subscriber not active")
+        if self.gps_sub is None:
+            self.get_logger().warning("GPS subscriber not active")
+        if self.global_path_sub is None:
+            self.get_logger().warning("Global Path subscriber not active")
+        if self.filtered_wind_sensor_sub is None:
+            self.get_logger().warning("Filtered Wind Sensor subscriber not active")
```
jamenkaye: Would it be an improvement to log all the inactive subscribers in the same message, eg
 `Inactive subscribers: AIS Ships, GPS, ...` ? This would help keep logs from getting cluttered.

jamenkaye (commented):
Nice work! I added one thing to consider 👍

local_pathfinding/node_navigate.py
Krithik1: I think its a good idea
I can change that

jamenkaye (approved):
Great implementation! If you could please change the warning messages to contain the exact topic names in [custom interfaces](https://github.com/UBCSailbot/custom_interfaces), this will be ready to get merged in!

You can go ahead and merge this as soon as you make the change :)

PR #88, opened 2024-02-14
Calculates current heading

### LINKED ISSUE #84: Publish desired heading
### Purpose

Currently we return 0.0 as the desired heading in `node_navigate.py` We need to return the heading of the next global waypoint from the current boat position.

### Description

[See above]

### Resources

-

### LINKED ISSUE #283: Publish desired heading
### Purpose

Currently we return 0.0 as the desired heading in `node_navigate.py` We need to return the heading of the next local waypoint from the current boat position.

### Description

[See above]

### Resources

- https://github.com/UBCSailbot/local_pathfinding/blob/main/local_pathfinding/node_navigate.py#L154

Krithik1:
### Description

- Resolves UBCSailbot/sailbot_workspace#283 
- Gets current waypoint from GPS subscriber and finds the waypoint in self.local_path.waypoints
- If found, calculates the heading based on formula linked in resources
- GPS data not being sent, so running into error. If mock GPS data then works fine

### Verification

- [x] Used mock GPS data for the current waypoint. Used the first value in the waypoints list.

### Resources

- [Link to formula](https://groups.google.com/g/google-earth-browser-plugin/c/5TKIdi7dxWc?pli=1)

### DIFF
--- a/local_pathfinding/node_navigate.py
+++ b/local_pathfinding/node_navigate.py
@@ -188,10 +188,10 @@ class Sailbot(Node):
         if next_waypoint is None:
             return 0.0
         heading = GEODESIC.inv(
-            current_waypoint.longitude,
-            current_waypoint.latitude,
-            next_waypoint.longitude,
-            next_waypoint.latitude,
+            lats1=current_waypoint.latitude,
+            lons1=current_waypoint.longitude,
+            lats2=next_waypoint.latitude,
+            lons2=next_waypoint.longitude,
         )[0]
         return heading
 
--- a/local_pathfinding/node_navigate.py
+++ b/local_pathfinding/node_navigate.py
@@ -1,12 +1,13 @@
 """The main node of the local_pathfinding package, represented by the `Sailbot` class."""
 
-import math
-
 import rclpy
+from pyproj import Geod
 from rclpy.node import Node
 
 import custom_interfaces.msg as ci
 from local_pathfinding.local_path import LocalPath
+
+GEODESIC = Geod(ellps="WGS84")
 
 
 def main(args=None):
@@ -177,22 +178,21 @@ class Sailbot(Node):
             self.gps, self.ais_ships, self.global_path, self.filtered_wind_sensor, self.planner
         )
 
-        # TODO: create function to compute the heading from current position to next local waypoint
-        # Find index of self.gps in current waypoints
         current_waypoint = self.gps.lat_lon
         next_waypoint = None
         waypoints = self.local_path.waypoints
         for i in range(len(waypoints)):
-            if waypoints[i] == current_waypoint and i != len(waypoints) - 1:
-                next_waypoint = waypoints[i+1]
+            if waypoints[i] == current_waypoint and i < len(waypoints) - 1:
+                next_waypoint = waypoints[i + 1]
                 break
         if next_waypoint is None:
             return 0.0
-        heading = math.atan2(
-            (next_waypoint.longitude-current_waypoint.longitude) *
-            math.cos(current_waypoint.latitude),
-            (next_waypoint.latitude-current_waypoint.latitude)
-        )
+        heading = GEODESIC.inv(
+            current_waypoint.longitude,
+            current_waypoint.latitude,
+            next_waypoint.longitude,
+            next_waypoint.latitude,
+        )[0]
         return heading
 
     def update_params(self):
--- a/local_pathfinding/node_navigate.py
+++ b/local_pathfinding/node_navigate.py
@@ -1,9 +1,11 @@
 """The main node of the local_pathfinding package, represented by the `Sailbot` class."""
 
-import custom_interfaces.msg as ci
+import math
+
 import rclpy
 from rclpy.node import Node
 
+import custom_interfaces.msg as ci
 from local_pathfinding.local_path import LocalPath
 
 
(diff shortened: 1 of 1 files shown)


local_pathfinding/node_navigate.py
```diff
@@ -176,7 +178,22 @@ def get_desired_heading(self) -> float:
...
+            if waypoints[i] == current_waypoint and i != len(waypoints) - 1:
+                next_waypoint = waypoints[i+1]
+                break
+        if next_waypoint is None:
+            return 0.0
+        heading = math.atan2(
+            (next_waypoint.longitude-current_waypoint.longitude) *
+            math.cos(current_waypoint.latitude),
+            (next_waypoint.latitude-current_waypoint.latitude)
+        )
```
jamenkaye: Mathematically this looks like it will work! You can improve the readability and accuracy by converting the latlons to XY using methods in `coord_systems.py`, and then do trigonometry on the XY components. You could also use `GEODESIC.inv()` method directly.

local_pathfinding/node_navigate.py
```diff
@@ -176,7 +178,22 @@ def get_desired_heading(self) -> float:
...
+        return heading
```
jamenkaye, line 196: I think it looks correct, but I just wanted to double check that when it gets published, `desired_heading` follows the conventions in custom interfaces? 
https://github.com/UBCSailbot/custom_interfaces/blob/main/msg/DesiredHeading.msg
https://github.com/UBCSailbot/custom_interfaces/blob/main/msg/HelperHeading.msg

local_pathfinding/node_navigate.py
```diff
@@ -176,7 +178,22 @@ def get_desired_heading(self) -> float:
         )
 
         # TODO: create function to compute the heading from current position to next local waypoint
-        return 0.0
+        # Find index of self.gps in current waypoints
+        current_waypoint = self.gps.lat_lon
+        next_waypoint = None
+        waypoints = self.local_path.waypoints
+        for i in range(len(waypoints)):
+            if waypoints[i] == current_waypoint and i != len(waypoints) - 1:
```
jamenkaye: I think this should be `i < len(waypoints) - 1`. 
Since `len(waypoints)` is the number of waypoints, but python uses zero indexing, the largest valid index is `len(waypoints) - 1`. The goal here is to ensure that `(i+ 1)` is valid, so we need that `i + 1 <= len(waypoints) - 1` , or in other words, `i < len(waypoints) - 1`

local_pathfinding/node_navigate.py
jamenkaye, line 196: Would it be possible to add some tests for this function?

jamenkaye (changes requested):
Good start! I left a few comments 👍

local_pathfinding/node_navigate.py
jamenkaye: Geodesic.inv [documentation](https://pyproj4.github.io/pyproj/stable/api/geod.html#pyproj.Geod.inv)

I'd suggest using named arguments, ie `geod.inv(lats1=10, ...)`, since the order of the arguments is kinda unintuitive. Normally I think latitude should come first (latitude, longitude), but their function flips this order.

You can also check `coord_systems.py` for example usage

local_pathfinding/node_navigate.py
Krithik1: For the tests, should I create a new test_node_navigate.py file and manually set the value of self.gps to calculate the desired heading?

local_pathfinding/node_navigate.py
```diff
@@ -175,8 +178,22 @@ def get_desired_heading(self) -> float:
             self.gps, self.ais_ships, self.global_path, self.filtered_wind_sensor, self.planner
         )
 
-        # TODO: create function to compute the heading from current position to next local waypoint
-        return 0.0
+        current_waypoint = self.gps.lat_lon
```
jamenkaye, line 181: I believe `self.gps.lat_lon` is the current GPS position of the boat, which is generally not one of the waypoints. I think Jay might be working on code to determine the next local waypoint, but we can sync up with him tomorrow

local_pathfinding/node_navigate.py
```diff
@@ -175,8 +178,22 @@ def get_desired_heading(self) -> float:
             self.gps, self.ais_ships, self.global_path, self.filtered_wind_sensor, self.planner
```
jamenkaye, line 178: This is about line 175:

The function description says we should return something that violates the heading convention in the case of an error, but the [heading convention](https://github.com/UBCSailbot/custom_interfaces/blob/main/msg/HelperHeading.msg) allows $-180 \degree < \text{heading} \leq 180 \degree$. 

Can you please change the return value on line 175 to something outside those bounds? I'd suggest 404 since that's a common error number for websites haha

local_pathfinding/node_navigate.py
```diff
@@ -175,8 +178,22 @@ def get_desired_heading(self) -> float:
...
+        current_waypoint = self.gps.lat_lon
+        next_waypoint = None
+        waypoints = self.local_path.waypoints
+        for i in range(len(waypoints)):
+            if waypoints[i] == current_waypoint and i < len(waypoints) - 1:
+                next_waypoint = waypoints[i + 1]
+                break
+        if next_waypoint is None:
+            return 0.0
+        heading = GEODESIC.inv(
```
jamenkaye, line 190: I think `GEODESIC.inv()` has multiple returns ([documentation](https://pyproj4.github.io/pyproj/stable/api/geod.html#pyproj.Geod.inv:~:text=or%20array%20%E2%80%93-,Forward%20azimuth(s),-(azi12))), and I think you only want the forward azimuth (image at bottom)

I believe Patrick is a fan of this kind of syntax for this situation: `_, value, _ = myfunc()`, where `myfunc()` returns a touple:

```
def myfunc():
	return (1, 2, 3)
```
In this example, `print(value)` outputs 2.


image

local_pathfinding/node_navigate.py
jamenkaye: I think this would be the best way to test this:
1. Create a helper function to `get_desired_heading()` that accepts "simple" arguments like waypoints or lists of waypoints, rather than "complicated" arguments like `self` (which is a whole sailbot object). `get_desired_heading()` will basically just call the helper function.
2. Create `test_node_navigate.py` like you suggested, and just test the helper function. I'd suggest using simple tests like when the desired heading is $0 \degree$ or $90 \degree$, and maybe some more complicated tests where the "actual" value is computed with geodesic.inv() as well.

jamenkaye (commented):
Hi Krithik, I left a few comments! We might need to add some more complexity here depending on if we need to determine which local waypoint we're going towards. Jay might be working on this tho so we can talk to him tomorrow.

Issue #89, opened 2024-02-21
Migrate Python Dependencies for ROS Specific Functionalities from Docker Image to package.xml


SPDonaghy:
### Purpose

We want to distinguish between dependencies that are required by our ROS nodes and those which are not.

For example: 

**NumPy** is used by some of our ROS nodes so we will specify that dependency in the package.xml file, ensuring that dependency is built into the package.

**Flask** is used for creating mock global paths and is not involved with any ROS processes, so the dependency is specified in the Docker image.

### Changes

Moving any dependencies, for ROS nodes in the local_pathfinding package, from the base-dev.Dockerfile to the package.xml file.

### Resources

- [Sailbot Workspace Parent Issue](https://github.com/UBCSailbot/sailbot_workspace/issues/242)

patrick-5546:
Just forgot to link to pr

PR #90, opened 2024-02-21
Migrate Local Pathfinding Python Dependencies to package.xml

SPDonaghy:
### Description

- Resolves to #89 

Added local_pathfinding Python dependencies to the package.xml file
I also removed these dependencies from the base-dev.Dockerfile in a separate SailbotWorkspace branch.

I did not remove NumPy from the Dockerfile, but I still added it to package.xml, which might be redundant but it does make it clear that this package specifically depends on NumPy.

The OMPL dependency is still defined in the base-dev.Dockerfile.

### Verification

1. I removed the Python dependencies from `base-dev.Dockerfile`, then built a new image based off of the new Dockerfile.
2. I modified the Dev Container Dockerfile to use this new image and rebuilt the container.
3. I ran the `setup` task and then the command `pip3 list`
4. The Python dependencies specifed both in `base-dev.Dockerfile`and in `local_pathfinding/package.xml` were all installed successfully

### DIFF
--- a/global_paths/path_builder/requirements.txt
+++ b/global_paths/path_builder/requirements.txt
@@ -1,2 +1,0 @@
-Flask 3.0.0
-Werkzeug 3.0.1
--- a/package.xml
+++ b/package.xml
@@ -21,11 +21,6 @@
   <test_depend>ament_flake8</test_depend>
   <test_depend>ament_pep257</test_depend>
   <test_depend>python3-pytest</test_depend>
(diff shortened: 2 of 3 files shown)


patrick-5546 (commented):
Can you create a sailbot workspace PR for your branch there? We can present this in the software-wide meeting tomorrow

patrick-5546 (approved):

